PR #130, opened 2014-11-18
Add NetCDF topography capability

mandli:
This PR develops the ability to read in topography data in a NetCDF file of the sort GEBCO uses.  I am not completely sure if this a universal format but it works on test data so far.  Currently the biggest question is whether we want to include the pre-processor directives in the file or have a duplicate `topo_module` source that includes the NetCDF.  I favor the former approach but am open to discussion.  It should be noted that by changing the file suffix to `F90` that gfortran should automatically call the pre-processor.


mandli:
This requires PR clawpack/clawutil#84 to pass the tests.

mandli:
This now should work and be robust enough to include in the master branch (hopefully the tests pass).  One caveat to this implementation is that it depends fairly heavily on the [NetCDF CF meta-data conventions](http://cfconventions.org).  The upshot is that the GEBCO data sets all use these conventions.  Other data sets can also be read into GeoClaw by reading the data set using `topotools` and the `nc_params` argument to the `read` function and then calling `write` which should put the file into the correct format.

mandli:
Tests are now passing again.  There was some trickiness with side-stepping the NetCDF tests if it was not available.

mandli:
Any objections, comments or questions about this PR?  I need this functionality for doing some global tidal work along with easier access to high-resolution bathymetry data bases.

src/2d/shallow/topo_module.f90
```diff
@@ -401,8 +404,12 @@ end subroutine set_topo_for_dtopo
 
     subroutine read_topo_file(mx,my,topo_type,fname,topo)
 
+#ifdef NETCDF
+        use netcdf
```
rjleveque: What is this idef checking?  I get an error when I try to compile in `examples/tsunami/chile2010` without setting any preprocessing flag.

```

        use netcdf
                  1
Fatal Error: Can't open module file 'netcdf.mod' for reading at (1): No such file or directory
```

examples/tsunami/chile2010/Makefile
```diff
@@ -23,7 +23,8 @@ RESTART = False
 # Environment variable FC should be set to fortran compiler, e.g. gfortran
 
 # Compiler flags can be specified here or set as an environment variable
-FFLAGS ?=
+FFLAGS = -O3 -funroll-loops -finline-functions -fopenmp -cpp -DNETCDF -I$(NETCDF4_DIR)/include
+LFLAGS = -fopenmp -lnetcdff
```
rjleveque: Presumably we should not have this netcdf dependency in the standard chile2010 example?

src/2d/shallow/topo_module.f90
mandli: It is just checking to see if `-DNETCDF` was set at compile time.  If not it should remove the line. By default the pre-processor flag should be added though, is -cpp really not being used?

examples/tsunami/chile2010/Makefile
mandli: Probably not, forgot to remove this.

src/2d/shallow/topo_module.f90
mjberger: -cpp flag ands some complications, it’s not a good thing to do in general, IMHo.

 — Marsha

mandli: I agree with you in general (especially when they are used too much) but there is no other way to do this cleanly in Fortran without duplicating a lot of source.  This depends on the pre-processor only a minimal ammount and we should try to keep any use of it that way.

rjleveque:
Ok, the chile2010 example now works for me (using the new Makefile without netcdf flag set).

PR #80, opened 2014-06-02
Merge Storm Surge Code Into Main Code

mandli:
In an effort to cut down on the upkeep of the storm surge code I decided to try and merge the storm surge specific code back into the general "shallow" code.  With a few modifications of what was being done with the storm surge code, the modifications were actually pretty minimal.  The primary bother for existing applications will be adding the required storm surge modules to the Makefiles.  Crude speed tests are below  but at least for the examples in GeoClaw do not show and performance degradation.

### Timings
`FC=gfortran (v4.8.2)`
`FFLAGS=-O3 -funroll-loops -finline-functions -fopenmp`
Results were averaged over 3 runs

Example   | Old Time | New Time
-------------|--------------|----------------
Chile2010   | 16.16900000 s | 14.64300000 s
bowl-slosh  | 3.18000000 s  | 3.00400000 s
bowl-radial | 10.74600000 s | 10.00000000 s
Ike         | 61.82000000 s | 60.14500000 s

PR #134, opened 2015-03-07
Topotypes 4 and 5 added

rjleveque:
WIP:  @mandli should comment.

New topo_types 4 and 5 are analogous to 2 and 3 but allow dx != dy, so the header has two lines, one for dx and one for dy instead of a single line with cellsize = dx = dy.

Modifications to both topotools.py and to the Fortran code processing the topo.  This was already supported for the most part in the Fortran code but we didn't have a topo type to read it in this way.  This modification was required for http://depts.washington.edu/clawpack/geoclaw/benchmarks/nthmp_currents_2015/problem3/

Changed `Topography.delta` to be a list `[dx,dy]` instead of a single `cellsize`.  For topo type 2 or 3 it is a list with equal numbers.  This might cause problems somewhere?

Also added some checks to make sure dx=dy assumption is respected when that can be checked.  As a result, found a bug in `tests/storm_surge` regression test.


mandli:
A few thoughts:
 - Can we continue to use `topo_type` 2 and 3 for 4 and 5 and not check the delta condition?  In python we can get around the backwards compatibility problem by checking to see how many dx are specified.
 - Furthermore on the topic of `topo_type` I would suggest that we only have different `topo_type` for truly different file formats.  Two in particular come to mind with NetCDF topography files and other widely available binary formats.
 - Thanks for finding the storm_surge problem!  I will raise another issue on that.

rjleveque:
Responses...

* In the Fortran code we also need to know whether to read in one dx or both dx,dy.  I'm sure this can be done if we stick with just 2 and 3, but I didn't give it much thought.  (It would also be nice if the Fortran code allowed the value and label to appear in either order on a line, as the Python now does, since topo downloaded from etopo1 for example has them in the opposite order and `topotools.swapheader`always has to be applied.)

* It would be good to allow a wider variety, particularly binary formats since these would be much smaller files.  

* Also MOST format can be handled by converting with `most2geoclaw` but it might be nice to make this a directly supported data type.

* I think I fixed the problem in storm_surge and updated the regression data in this PR.

mandli:
So should we go ahead and merge this or think more on whether we want to add more `topo_type`s and then possibly change them in the future?  I have a branch somewhere that adds NetCDF reading of data which could also be added to this.

rjleveque:
Let's hold off for now since it's not yet documented and we might want to restructure.  We can discuss in SLC next week.

mandli:
Storm surge bug fix was merged into PR #80.

rjleveque:
Closing this in favor of #137, but we should make issues for some of the suggestions above.

PR #135, opened 2015-03-09
Remove test in src/2d/shallow/filval.f90 that xl==xl2 and yb==yb2 

rjleveque:
Remove test in src/2d/shallow/filval.f90 that xl==xl2 and yb==yb2 -- originally for debugging in @ca673af7d and uses absolute error in way that can fail, as discovered by a user,
 https://groups.google.com/forum/?fromgroups#!topic/claw-users/NO0m1UAczw4

### DIFF
--- a/src/2d/shallow/filval.f90
+++ b/src/2d/shallow/filval.f90
@@ -52,7 +52,6 @@ subroutine filval(val, mitot, mjtot, dx, dy, level, time,  mic, &
     ! External function definitions
     real(kind=8) :: get_max_speed
 
-    real(kind=8) :: xl2,yb2
 
     refinement_ratio_x = intratx(level-1)
     refinement_ratio_y = intraty(level-1)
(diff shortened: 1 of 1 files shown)


mjberger:
I thought I had changed all routines to compute the grid point locations in exactly  the same way so that a real equality test was possible. Maybe  the better way to fix it is to find the one that didn’t do it right, leading to the failed test, rather than changing this test to a relative test, or removing it altogether?

Marsha

rjleveque:
The two values that don't agree (from the email linked above) are:
3433973.3333333330        3433973.3333333335 
They disagree in the 17th digit. Both values yb and yb1 are computed in this routine but in different ways, so I don't think this indicates a bug.

mjberger:
and they are not actually computed the same way - it’s an old piece of code from before I used integer index spaces, so it would be a big job to convert. Oh well.

mandli:
Can we merge this in then?

mjberger:
i would say yes

PR #137, opened 2015-03-24
[WIP] Topotypes 2 and 3 - parse values to allow dx and dy on cellsize line

rjleveque:
Replaces #134.

Allows dx and dy on the same line for topotype 2,3 rather than introducing new topotypes. 

Also allows values or labels to come first, so `topotools.swapheader` does not need to be applied to data from NGDC.  Not yet fully tested.

The new subroutine `parse_values` in `utility_module.f90` might be useful more generally.  Contains ugly string manipulation in Fortran -- maybe someone knows a better way to write this...

@mandli: I haven't yet rewritten the Python tools yet to support this.  Is it ok to change `Topography.delta` to always be a tuple `(dx,dy)` as in #134?

### DIFF
--- a/src/2d/shallow/test_parse_values.f90
+++ b/src/2d/shallow/test_parse_values.f90
@@ -1,0 +1,42 @@
+
+
+program test_parse_values
+
+    ! To test: select a test below and then:
+    !    gfortran utility_module.f90 test_parse_values.f90 
+    !    ./a.out
+
+    use utility_module, only: parse_values
+    use topo_module, only: read_topo_header
+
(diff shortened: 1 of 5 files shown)


mandli:
I am fine with changing `delta` to always be a tuple, that actually makes it more consistent with the solutions.

I also just restarted the Travis tests, they failed due to a timeout which is weird.

Also, did you want to include the test program?

mandli:
Now it looks like some of the tests need to be updated to maybe support the new `delta`.

Issue #139, opened 2015-04-13
Duplicate Symbol Definitions for `setbestsrc`


mandli:
New functionality in AMRClaw that includes gauge output I think has broken GeoClaw.  The `gauge_module` in AMRClaw now includes the `setbestsrc` subroutine as does `dumpgauge.f` so a conflict emerges when `use gauge_module` is done in `setbestsrc` in GeoClaw.  Relevant compile time error:
```
clawpack/geoclaw/src/2d/shallow/dumpgauge.f:140.9:

      use gauges_module                                                 
         1
Error: 'setbestsrc' of module 'gauges_module', imported at (1), is also the name of the current program unit
make: *** [/clawpack/geoclaw/src/2d/shallow/dumpgauge.o] Error 1
make: *** [new] Error 2
```
If `setbestsrc` is commented out in GeoClaw then we also have a problem since `tick` uses this and does not include the `gauge_module` so we arrive at undefined symbols
```
Undefined symbols for architecture x86_64:
  "_setbestsrc_", referenced from:
      _tick_ in tick.o
ld: symbol(s) not found for architecture x86_64
collect2: error: ld returned 1 exit status
make: *** [xgeoclaw] Error 1
```
So the solution is probably just to insert a `use gauge_module, only: setbestsrc` in `tick` or something similar and remove `setbestsrc` from `dumpgauge.f`.

rjleveque:
The `setbestsrc` routine in the new `gauges_module.f90` is not the same as in the old `dumpgauge`, so instead see #140.

PR #140, opened 2015-04-19
Gauges module

rjleveque:
Fix issue #139 by copying `clawpack/amrclaw/src/2d/gauges_module.f90` to `src/2d/shallow` and modifying to plot `h,hu,hv,eta` and check for dry neighbors.

Modified Makefiles to point to new version of this module and no longer use `dumpgauge.f`.

@mandli:  I did not modify the `Makefile` in `examples/multi-layer/plane_wave` since this uses a different version of `dumpgauge`.  Also some of the other modules used in this Makefile have been moved to `src/2d/shallow/surge` and this needs fixing.

Also `tests/multilayer/Makefile` has not been changed.


mandli:
I can fix the multi-layer Makefile as it needs updating anyway.  It looks like this is failing though, is this because of dependence on changes to AMRClaw as well?

mandli:
Finally figured out what was failing:
```
/home/travis/build/clawpack/geoclaw/clawpack/geoclaw/src/2d/shallow/advanc.f:230.29:
           call print_gauges(alloc(locnew),alloc(locaux),xlow,ylow,     
                             1
Error: Element of assumed-shaped or pointer array passed to array dummy argument 'q' at (1)
```
I think this is cropping up because the function interface is no longer implicitly declared.  I am not sure why `gfortran` picks it up on travis but not on our machines.  Maybe @mjberger can test this with the intel compiler?

Also, something related here:  

http://coding.derkeiler.com/Archive/Fortran/comp.lang.fortran/2013-02/msg00125.html

mandli:
I did some testing using the intel compiler and think replacing the statement in question with
```fortran
           call print_gauges(alloc(locnew:mitot*nvar*mjtot), 
     .                       alloc(locaux:mitot*nvar*mjtot),xlow,ylow,
     .                       nvar,mitot,mjtot,naux,mptr)
```
should work.

rjleveque:
This seems to fix it.  

Presumably the same change should be made in clawpack/amrclaw?  Why didn't travis complain there, e.g. in https://github.com/clawpack/amrclaw/pull/118?

mandli:
I am not sure actually.  I just tested the acoustics_2d_radial problem and it did not complain (with the intel compiler).  Do we have an example that uses gauges in AMRClaw?

rjleveque:
I think all the regression tests in the `amrclaw/tests` directory use gauges to test results.

mandli:
Huh, well I am not sure then.  I would say we merge this, should we make the same change in AMRClaw even though it's not failing then?

rjleveque:
OK, I'll make the same change there and merge this.

rjleveque:
@mandli -- I just looked at this code more carefully when putting into amrclaw in 3d -- is it right?

Shouldn't it be:

```
           call print_gauges(alloc(locnew:locnew + mitot*nvar*mjtot), 
     .                       alloc(locaux:locaux + mitot*nvar*mjtot),xlow,ylow,
     .                       nvar,mitot,mjtot,naux,mptr)
```
i.e. after the colon should be the ending index, not the number of elements?

mandli:
I think you are right although apparently it is only passing the index into the array (which is odd).  Perhaps because assumedly `locnew > mitot * nvar * mjtot` the compiler does not check for the sanity of the indexing.

mjberger:
are you sure it doesn’t copy the subarray, or have any hidden side effects?

Marsha

mjberger:
also, this is used all over, in the old code.  Maybe there’s something else triggering it, if this is the only one that complains.

Marsha

mandli:
I think there are two issues here:
1. The subroutine interfaces are explicit if in a module so the compilers CAN check to make sure the arguments are consistent which I think is what triggered this (very inconsistently though).
2. The discrepancy between dummy arguments and actual arguments is one that is riddled with special cases.  I learned even more when reading one of the links in the above (discussion)[https://software.intel.com/en-us/blogs/2009/03/31/doctor-fortran-in-ive-come-here-for-an-argument] which attempts to clarify when each is used.  In particular the explanation of "sequence arguments" was applicable.
3. The intel compiler actually provides a bit more detail about the problem that would be worth thinking about although I am not sure how it applies here yet:

> If the actual argument is scalar, the dummy argument shall be scalar unless the
> actual argument is of type character or is an element of an array that is not
> assumed shape, pointer, or polymorphic.

PR #150, opened 2015-05-07
Googleearth

donnacalhoun:
Several examples for the Google Earth Visualizations : 

* example/tsunami/chile2010

* storm/Ike  (see setplot_ge.py)

* examples/tsunami/LaPush (not a complete example, since there is not a setrun file). May this should go in apps?

* Additions to kmltools.py


donnacalhoun:
I noticed something odd in the storm surge simulations.  Does anyone know why there are gaps between amr grids here?  I didnt' find anything obvious. 

storm_cropped

mandli:
Does that happen with the tsunami case as well?  It might always happen and I have never noticed since I have a different colormap than is used in most of the tsunami simulations.

mandli:
Couple of questions/comments:
 - Why did the `.gitignore` get deleted?
 - Do we want to add LaPush as an example here or put it in the apps repository?
 - There are a bunch of changes to the Chile 2010 examples plots, did you intend to have a separate `setplot_ge.py` like for Ike?
 - I don't think the LaPush example is going to compile since it is missing module source required now in the newest versions of GeoClaw.

donnacalhoun:
I haven't see this with other simulations.   But I wonder if has something to do with `contourf` - maybe there is a lower threshold value?  I noticed you had 

        cmap._rgba_under = (0.0,0.0,0.0,0.0)

which I set to `(0,0,0,1)`, but this didn't do anything, and I didn't pursue it further.

mandli:
I am pretty certain is the `contourf` plotting, I just have not done any tsunami plotting with `contourf`.  Using `pcolor` for the surge plotting unfortunately takes forever which is why I switched to `contourf`.

donnacalhoun:
* I don't think I deleted `.gitignore`.    But in general I don't track them, so hopefully, I didn't do something inadvertently. 

* I can add a setplot_ge.py for Chile, or revert to the original and add  the KML stuff.    This example wasn't quite as complex at the Ike, so I thought it was okay to keep the KML in original `setplot.py` file.  But we had been working on this file for so long,that probably other stuff changed as well. 

* Early on, we added LaPush under version control, just to be able to work on it together.  But I only ever had a few output files that Randy sent a long time ago (and no setrun.py file).  It is a nice example, since it shows several zoomed figures.   We'd have to rely on Randy, though to get a complete example.  working. 

* There is also a Chile 2014 example, which I don't have under version control.

donnacalhoun:
Does visclaw use pcolormesh?   This is quite a bit faster, I've found.

mandli:
We do support `imshow` which is even faster than `pcolormesh` but unfortunately it does not always play nicely with AMR layering (neither does `pcolormesh`) and land.  `contourf` is also faster than either of these for the storm-surge plotting.

mandli:
- You deleted the entirety of the `.gitignore` file
 - There were a bunch of changes beyond the KML additions to the Chile example.  Also, since the generation of the tiles takes sometime I think I would prefer to either turn the KML generation off by default but keep it in the same file.
 - @rjleveque should we include LaPush?
 - I would not worry about the Chile 2014 example unless you want to put in the apps repository

donnacalhoun:
Which .gitignore did I delete?  Was it yours?  I can add it back in, if you let me know which one it was.

mandli:
It's the one in the base of the repository, look at the files changed tab above.

donnacalhoun:
Yes - KML should be turned off by default.  I'll revert to the original Chile and add the KML stuff back in.

rjleveque:
- I think kml should be off by default 
- Also tiling should be off by default when kml is on, since it requires gdal?
- Other changes in `chile2010/setplot.py` should be reverted if possible to match `master`.
- We should not include LaPush since this example doesn't include `setrun.py` or all the required topo data.  This is WIP and not ready for public consumption.  But I will try to put together some example that includes inundation to better illustrate both kml and fgmax.  This will go in clawpack/apps since it will require a longer run time than we like for the core examples.

I'll take a closer look at this PR later today....

rjleveque:
If you try to open the file `_plots/Chile_2010.kml` it just sort of hangs since it points to a non-existent webpage by default.  Users might think this is the file to open.  Should the default be to not produce this file?

donnacalhoun:
Good point.  I put in a fake URL, which doesn't make much sense, actually.   The default is None, in which case, the KML file is not created.

donnacalhoun:
I commented out the publish URL.   I can also just remove it, but at least if I leave it in, users will know about it.  In any case, I'll include something about this in the documentation.

rjleveque:
In `examples/tsunami/chile2010/setplot.py`, the default has `plotdata.kml = False` but it still produces figures 1 and 2 and adds them to  `_plots/_PlotIndex.html`.  Shouldn't it ignore figures with `plotfigure.use_for_kml = True` when `plotdata.kml = False`, or does the user also have to set `plotfigure.show = False`?   If so, then the default `setplot.py` should have these set to `False`.

donnacalhoun:
Good point.  So if `plotdata.kml = False`, I'll ignore any `plotdata.use_for_kml` settings and not create the plot.   I might have been following what happens if  `plotdata.html = False`.  In this case, plots might still be created (for Latex output, for example), right?   This wouldn't make much sense for the KML plots, though.

But in fact, have wondered all along whether the KML plots should be in the PlotIndex file at all.

rjleveque:
I guess I'm still in favor of having a different `setplot_kml.py` in this example directory for the kml plots.  In general it makes sense to have everything in one file, but currently this is the first example geoclaw users look at (and the only real tsunami example in the `examples` directory), and I most users will not want to do kml plots right off.  Adding all the kml stuff makes `setplot.py` much more intimidating for new users, IMHO.

donnacalhoun:
I think this a good idea.  I'll create a separate `setplot_kml.py` file.

rjleveque:
In general it probably doesn't make sense for the kml plots to show up in _PlotIndex.html.

donnacalhoun:
It seems that visclaw removes all figure files from the plot directory before creating new figure files.  Is that right?  So if figures 1,2,3 have been created for visualization in HTML, and then figures 4,5 are 
created for visualization in KML, the figures 1,2,3 will be deleted before the figures 4,5 are created.   This would make it hard to easily switch between visualizations.

rjleveque:
That's true.  I guess I would handle this one of two ways in my own workflow:

- Create the kml plots, move `_plots` to `_plots_kml`, and then create the html plots.

- Or more cleanly, if I was doing this a lot, I'd probably create a `Makefile_kml` that has

```
    SETPLOT_FILE = setplot_kml.py     
    PLOTDIR = _plots_kml
```
so that `make plots -f Makefile_kml` uses the proper `setplot` file and creates a second plots directory.

rjleveque:
I suggest a different default colormap for kml, e.g.

```
    transparent = [0.0, 0.0, 0.0,0.0]
    red_a = [1.0,0.0,0.0,1.0]
    light_green_a = [0.0,1.0,1.0,1.0];
    cmap = colormaps.make_colormap({-1.0:light_green_a,
                                                0.0:transparent,
                                                1.0:red_a})
```
The light_green is the color we use for the undisturbed ocean in the standard geoclaw tsunami plots, and shows up much better against the dark blue ocean on Google Earth instead of using dark blue for surface elevations < 0.

donnacalhoun:
Sounds like a good idea.  Did you also notice that the computational domain (under regions) has a solid background, making it easier to see the tsunami disturbance?  

The colormap with the light blue (instead of transparent) can also make it easier to see the wave.

rjleveque:
Merging this, along with clawpack/visclaw#150.

PR #151, opened 2015-05-11
change order of operations in tick.f so restart generates exactly same grids

mjberger:
…t restart matches


rjleveque:
Works for me -- I did several tests in chile2010.

rjleveque:
Goes with clawpack/amrclaw#133.

PR #155, opened 2015-05-17
Add binary output for multi-layer shallow water equations

mandli:


mandli:
This can wait until after 5.3.0 as it is a new feature and is not critical.

PR #156, opened 2015-05-17
Gauge change

mjberger:
add restart capability to gauges in geoclaw, by passing the logical variable 'rest' to set_gauges.


rjleveque:
This breaks several regression tests.  I found the problem in two of them -- bugs in the tests since `tests/dtopo1/Makefile` and `tests/storm_surge/Makefile` point to `  $(AMRLIB)/amr2.f90` and should point to `$(GEOLIB)/amr2.f90`.  

There is still a problem with `tests/multilayer`  -- the gauge output expected by the regression test is not generated.  Probably because `src/2d/shallow/multilayer/gauges_module.f90` needs to be updated in the same way as done to `src/2d/shallow/gauges_module.f90`

@mjberger: can you make these changes and push them?

rjleveque:
I fixed to two Makefiles in #157.  

I think this change to gauges on restart can wait for the next release -- we should make the same change in amrclaw and do some more testing, documenting, etc. and then include it in some new examples showing how to do restarts.

I anticipate a v5.3.1 before long to fix various things if nothing else.

mandli:
@mjberger can you merge in the current master branch so that the tests can be run?

mandli:
The current master branch now includes a fix for the tests so you need to pull those changes into this branch for the testing to work again.  You could do
```
git fetch clawpack
git merge clawpack/master
```
if clawpack is the remote pointing to the clawpack organization's repository.  Hopefully there are not any conflicts that will be difficult to deal with.

mandli:
It was mostly specifically for this PR (https://github.com/clawpack/geoclaw/pull/156).

rjleveque:
Note: see the discussion in clawpack/amrclaw#137 before merging this.

rjleveque:
Replaced by clawpack/amrclaw#144.

PR #162, opened 2015-07-10
Test GeoClaw run without matplotlib install

mandli:
This removes most of the dependency on matplotlib from GeoClaw and "skips" the tests that require it.  This includes only the comparison with the legacy `topotools` module.


mandli:
This was included in the new testing framework.

PR #166, opened 2015-07-25
dtopotools.py fixes

rjleveque:
standard_units was being overwritten and output format for writing subdivided fault fixed

### DIFF
--- a/src/python/geoclaw/dtopotools.py
+++ b/src/python/geoclaw/dtopotools.py
@@ -585,7 +585,8 @@ class Fault(object):
 
     """
 
-    def __init__(self, subfaults=None, input_units={}):
+    def __init__(self, subfaults=None, input_units={},
+                 coordinate_specification=None):
         r"""Fault initialization routine.
         
         See :class:`Fault` for more info.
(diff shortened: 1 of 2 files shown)


mandli:
I thought I had fixed this but the test `test_against_old` imports the old `topotools` which uses a `matplotlib` function for the test so we either need to put it in a try-except block similar to what's done for `scipy` or put it in the group of other tests which are not run without a command line argument.  I kind of prefer the earlier for all of the `matplotlib` tests but that would be a bit more work.

rjleveque:
Which test is failing?

mandli:
Not sure why it's failing here only.
```
ERROR: Test against the old topotools from 5.1.0.
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/build/clawpack/geoclaw/tests/test_topotools.py", line 132, in test_against_old
    X, Y, Z = old_topotools.topofile2griddata(file_path, topotype=1)
  File "/home/travis/build/clawpack/geoclaw/tests/old_topotools.py", line 94, in topofile2griddata
    import pylab
ImportError: No module named pylab
```

rjleveque:
I removed the pylab dependency and now travis is happy.

PR #167, opened 2015-08-23
CPU Timing

ortegagingrich:
This proposed merge updates the geoclaw timing and timing output to match the recently merged changes in the amrclaw repository.


mandli:
Does this work on your machine?  I am not sure what these tests have in common that would cause a failure.

rjleveque:
This works fine on my laptop but I just reran Travis on this PR and it still fails.  I'll try doing a test PR with a trivial change from master and see if that fails.

mandli:
Just restarted the tests to make sure this is now working with the update to clawpack/clawpack.

PR #171, opened 2015-09-27
bc2amr calling sequence wrong in amrclaw routines

rjleveque:
I just found a major bug that apparently shows up infrequently but sometimes generates garbage at the boundaries where the ocean should be flat.

The routine `bc2amr.f` in amrclaw has a different calling sequence than in geoclaw.  (Why??  I seem to recall we noticed this before but can't find an issue on it.)

The two routines `bound.f90` and `saveqc.f` both call `bc2amr.f`.  Currently geoclaw application Makefiles point to the amrclaw version of these, which then call the geoclaw version of `bc2amr` with the wrong calling sequence.  

For now I copied these to geoclaw and changed the calling sequence, and point to the new versions in Makefiles where this is a problem.  But ideally we would make the calling sequences identical so that the amrclaw versions of these routines can still be used rather than duplicating code.

So don't merge this yet, let's discuss how to proceed.

### DIFF
--- a/src/2d/shallow/bound.f90
+++ b/src/2d/shallow/bound.f90
@@ -1,0 +1,128 @@
+!
+!  :::::::::::::: BOUND :::::::::::::::::::::::::::::::::::::::::::
+!     This routine sets the boundary values for a given grid 
+!     at level level.
+!     We are setting the values for a strip ng zones wide all
+!     the way around the border, in 4 rectangular strips.
+!
+!     Outputs from this routine:
+!     The values around the border of the grid are inserted
+!     directly into the enlarged valbig array.
+!
+!     This routine calls the routine filpatch
+!     which for any block of mesh points on a given level,
+!     intersects that block with all grids on that level and with
+!     the physical boundaries, copies the values into the
+!     appropriate intersecting regions, and interpolates the remaining
+!     cells from coarser grids as required.
+! :::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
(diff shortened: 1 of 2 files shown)


mandli:
Given your second suggestion should we actually make this change in AMRClaw instead?

rjleveque:
yes, if it's true that we can simplify the calling sequence in the amrclaw version then that's the place to make changes.

But another question for @mjberger: Is there a reason `xperiodic, yperiodic,spheredom` are still in the calling sequences (in both versions)?  Like `xlower, ylower,xupper,yupper`, these are all in `amr_module.f90` (with slightly different names `xperdom, yperdom, spheredom`).

mjberger:
I have looked through my previous claw packs, but can’t find where it diverged. I do remember discussing taking it out of the calling sequence and using the module vars though.

mandli:
I would support changing the AMRClaw call.  This does bring up the fact though that this would be less of an issue with the proposed Makefile changes in clawpack/clawutil#88.

mandli:
@rjleveque I don't remember now where this stands, can this be merged in with the correct changes?

rjleveque:
We decided to modify the calling sequence in amrclaw and then also in geoclaw to remove all the parameters that are in the `amr_module`.  I haven't gotten to this yet.  We can do when I'm in NY if you or @mjberger don't get to it first.

rjleveque:
This has been addressed in clawpack/amrclaw#154 and clawpack/geoclaw#189, and should be available in 5.4.0.

PR #173, opened 2015-10-15
Faster bound

mjberger:
as the name says, and is sometimes substantially faster for large numbers of grids.  Has to go with matching amrclaw fasterBound branch.  Also has lots of other tiny fixes, mostly format statement enlargements, to also go with larger numbers of grids


rjleveque:
The problem is that this was branched off of `mjberger/master`, which is no longer up to date with `master`.

I tried merging `master` in and only had to fix merge conflicts in `src/2d/shallow/advanc.f` (some rearrangement of timing calls) to get it to merge.

The unit tests then pass on my computer, but won't on travis because this PR depends on clawpack/amrclaw#149, so that should be fixed and merged in first perhaps, or better yet before doing that merge, do a PR on clawpack/clawpack first that includes both these PRs so they get tested together before merging either.  See http://www.clawpack.org/developers.html#top-level-pull-requests

examples/storm-surge/ike/setplot.py
```diff
@@ -161,10 +162,10 @@ def gulf_after_axes(cd):
     plotaxes.ylimits = gulf_ylimits
     plotaxes.afteraxes = gulf_after_axes
 
-    surge.add_surface_elevation(plotaxes, plot_type='contourf', 
+    surge.plot.add_surface_elevation(plotaxes, plot_type='contourf',
```
rjleveque: This line is wrong -- should be `surgeplot.add_surface_elevation(...`

@mandli: not sure if this is wrong in other branches.

mandli: At some point there was a PR that replaced these updates and I have not had a chance to go back through and sync everything up.  Let me know if you find any other issues with the data objects as I imagine there are more.

rjleveque:
After consulting with @mandli and @mjberger, decided this is ready to merge.

PR #177, opened 2015-10-31
Add surge indexing options and modify surge refinement settings

mandli:
Allows setting empty arrays or None values for wind_refine and R_refine
parameters that control storm based refinement.  Also adds ability to
set the wind_index and pressure_index so that custom layout of the aux
array is allowed.


mandli:
@mjberger You were the one who suggested these modifications so if you could make sure I interpreted your requests correctly.

PR #178, opened 2015-11-03
Adapt `topotools.Topography.plot` to new `add_colorbar`

mandli:
Now a custom Normalize object is used to map from data to the colormap
to handle uneven topo limits when necessary (otherwise just uses the
standard linear mapping).

See clawpack/visclaw#118 for the original issue and clawpack/visclaw#169 for the solution.

### DIFF
--- a/src/python/geoclaw/topotools.py
+++ b/src/python/geoclaw/topotools.py
@@ -854,6 +854,8 @@ class Topography(object):
         """
 
         import matplotlib.pyplot as plt
+        import matplotlib.colors as colors
+
         import clawpack.visclaw.colormaps as colormaps
 
         # Create axes if needed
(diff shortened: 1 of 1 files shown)


mandli:
@mjberger Can you see if this fixes your plotting problems in the South China Sea?

rjleveque:
Seems to work well in a couple tests, even if limits are highly skewed toward water or land, e.g. in `$CLAW/geoclaw/scratch`:

```
from clawpack.geoclaw import topotools
topo=topotools.Topography()
topo.read('etopo10min120W60W60S0S.asc',2)
topo.plot(limits=(-6000,1))
topo.plot(limits=(-1,6000))
```

PR #179, opened 2015-11-06
fix delta to tuple in topotools write and plot functions

rjleveque:
Fixed a couple places that were missed in 377b7041ad

### DIFF
--- a/src/python/geoclaw/topotools.py
+++ b/src/python/geoclaw/topotools.py
@@ -777,8 +777,6 @@ class Topography(object):
                     outfile.write("%s %s %s\n" % (self.x[i], self.y[i], topo))
 
             elif topo_type == 1:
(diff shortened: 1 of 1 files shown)


mandli:
The test is complaining about the modification that will use a list of `delta`s.  This was to support that ability more broadly right?  Has this been implemented fully in this case?

rjleveque:
Travis behavior is strange -- making the topo works fine on my laptop but tests are failing for a different reason:  we forgot to update tests after merging in clawpack/clawutil#91 so the gauges are not being compared properly.  I'll work on fixing that first and then update this PR.

rjleveque:
Travis tests now passing after reverting clawpack/clawutil#91 in clawpack/clawutil#94.

I think it now fully supports `delta` always being a tuple, but i might have missed something.

PR #184, opened 2016-01-09
Set omp private variable time in gfixup.f

rjleveque:
This bug never showed up since time isn't used, but I'm working on a time-dependent variable sea level option that uncovered it.

I don't know that time really needs to be a private variable in this loop, but this was easiest way to fix.


mjberger:
odd that it just showed up, since there is a default(none).  Wonder what’s going on there.

rjleveque:
Why would it show up if `time` was never used?  It creates a private variable `time` but doesn't initialize it, same as the other private variables that were already set in the loop.

rjleveque:
Travis tests pass after modifying call to `genfromtxt` in `dtopotools` -- NumPy changed and `skiprows` is deprecated.

PR #186, opened 2016-01-13
Adopt new Makefile format for ml

mandli:


mandli:
Do not merge this yet, I need to modify the test code too.

PR #192, opened 2016-02-07
Batch gauge output

mjberger:
default array size set to 1000 per gauge. gauge data is output when array is full, or at each checkpoint. when restart just add to file. each thread assigned a different unit name. Files names gaugeXXXXX.txt


mandli:
Note that the tests are failing as the changes to how we write out gauges still needs the implementation of reading the gauges in Python.

rjleveque:
Obsolete, included in #208.

PR #193, opened 2016-02-15
Update multilayer example Makefile

mandli:
This updates to the new version of the Makefiles


rjleveque:
It runs ok although it gives a warning:
```
Makefile:48: Extraneous text after `else' directive
```

The original `Makefile` does too.

mandli:
Yeah, it's a strange leftover from older Makefile syntax that sometimes gets triggers (and sometimes causes failures completely).  I will try to find a solution to this.

mandli:
@rjleveque I think I fixed this.  Let me know if you still see the warning.

rjleveque:
That fixed it.

PR #196, opened 2016-03-17
fixes to make checkpt_style < 0 work in geoclaw

rjleveque:
I think this fixes the problem, but there might be something else I missed.  I haven't tested much yet.


mandli:
What does a negative checkpoint style enable?

rjleveque:
An apparently undocumented feature introduced in 5.3.1, see #197

PR #202, opened 2016-04-03
Topography header order

rjleveque:
add option to Topography.write to write header with ARCGIS order of label then value.

To make it possible to output newly created topo file in a way that can be read into ARCGIS.

### DIFF
--- a/src/python/geoclaw/topotools.py
+++ b/src/python/geoclaw/topotools.py
@@ -788,13 +788,23 @@ class Topography(object):
             
         return num_cells
 
-    def write(self, path, no_data_value=None, topo_type=None, masked=True):
+    def write(self, path, no_data_value=None, topo_type=None, masked=True, 
+                header_order=1):
         r"""Write out a topography file to path of type *topo_type*.
 
         Writes out a topography file of topo type specified with *topo_type* or
         inferred from the output file's extension, defaulting to 3, to path
         from data in Z.  The rest of the arguments are used to write the header
         data.
+
+        :Input:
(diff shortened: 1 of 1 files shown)


src/python/geoclaw/topotools.py
```diff
@@ -796,6 +797,15 @@ def write(self, path, no_data_value=None, topo_type=None, masked=True):
...
         data.
 
+        :Input:
+         - *path* (str)  - file to write
+         - *no_data_value* - values used to indicate missing data
+         - *topo_type* (int) - GeoClaw format topo_type 
+         - *masked* (bool) - unused??
+         - *header_order* (int) - indicates format of header lines
+             1 ==> write value then label  (original Fortran assumption)
+             2 ==> write label then value  (needed for .asc files in ArcGIS)
```
mandli: Can I suggest we make this a string parameter just in case there are other headers we may want to use beyond ARCGIS and what we have been doing up until now?

rjleveque:
You mean have options  `header_order = "geoclaw"` and `header_order = "arcgis"` for now?

mandli:
Yeah, or even label the usual one `"default"` maybe?  Come to think of it though GeoClaw does not handle the other does it?

rjleveque:
Now `header_style` can be 'geoclaw' or 'default' for value first,  'arcgis' or 'asc' for label first.

Both the Python and Fortran code will read files with either first, but if we add other options to `header_style` down the road, we will need to add more options for reading and writing.

PR #210, opened 2016-04-25
WIP: Topotool updates

rjleveque:
Some change in default behavior of `Topography.write` that should be examined.

Enhancement of `Topography.crop` to support coarsening a topo grid as well.

Several Travis tests currently fail, also on master, so we need to sort this out before merging.


mandli:
This seems odd that this is failing like it is.  Do we just need to update the tests?

rjleveque:
@mandli: Don't merge yet, I need to fix some things in topotools.

By the way, the file `tests/bowl_slosh/regression_data/gauge00001.txt` seems to have a different format than the gauge files now being generated.

rjleveque:
@mandli: I fixed up the regression tests that were broken here and travis is now happy, but I think there's a potential problem with the way gauge tests are done to be addressed at some point.  See clawpack/clawutil#101.

I changed the default `Topography.write` format to `%15.7e` since my earlier choice of `%9.3f` seemed potentially dangerous in retrospect.

mandli:
This all looks good to me.  Is this ready to be merged then?

rjleveque:
Yes, I think so.

PR #219, opened 2016-07-28
WIP: Implement Gauge Output Options

mandli:
Adds ability to specify which fields from both q and aux should be output
along with specification of the format per gauge.  Also includes a
refactor of the routines, the addition of a gauge_type, and a few other
changes.


mandli:
I wanted to get some feedback on the capabilities in this PR.  Note that the capability to control these new parameters has yet to be added and are currently hardcoded.  This is looking towards being able to output the storm fields at the gauges (aux array variables) hence that capability.

Couple of questions:
 - How should we specify these extra parameters without dropping backwards compatibility?  I was thinking we can add them to the end of the current gauge arrays.
 - Is it really a good idea to have the eta field in between q and aux fields?  Should we even output eta?  Should outputting eta be a special parameter in GeoClaw?

Lots more probably.

mandli:
I tried out two different ways for specifying gauges now:

1. Specify the index into the q/aux arrays that you want output
2. Specify a boolean array the length of `num_eqn`/`num_aux`

The second method is cleaner in the Fortran and more easily allows things like "all" and "none" to be implemented.  The one drawback was that this method required that the `GaugeData` know what `num_eqn` and `num_aux` was.  Also added was the future ability to output in a different format (not implemented though).

Other PRs that go along with this one:
 - https://github.com/clawpack/amrclaw/pull/165
 - https://github.com/clawpack/clawutil/pull/103

mandli:
Just as an idea of how we would specify this in a setrun.py file the first spec looks like
```python
    rundata.gaugedata.gauges.append([1, -95.04, 29.07, rundata.clawdata.t0, rundata.clawdata.tfinal])
    rundata.gaugedata.gauges.append([2, -94.71, 29.28, rundata.clawdata.t0, rundata.clawdata.tfinal])
    rundata.gaugedata.gauges.append([3, -94.39, 29.49, rundata.clawdata.t0, rundata.clawdata.tfinal])
    rundata.gaugedata.gauges.append([4, -94.13, 29.58, rundata.clawdata.t0, rundata.clawdata.tfinal])
    rundata.gaugedata.display_format = {1:"e10.2"}
    rundata.gaugedata.file_format = 'ascii'
    rundata.gaugedata.q_out_fields = {1:[1, 2], 
                                      2:[2]}
    rundata.gaugedata.aux_out_fields = {1:[5, 6, 7], 
                                        2:[7],
                                        3:[1]}
```
and the second like
```python
    rundata.gaugedata.gauges.append([1, -95.04, 29.07, rundata.clawdata.t0, rundata.clawdata.tfinal])
    rundata.gaugedata.gauges.append([2, -94.71, 29.28, rundata.clawdata.t0, rundata.clawdata.tfinal])
    rundata.gaugedata.gauges.append([3, -94.39, 29.49, rundata.clawdata.t0, rundata.clawdata.tfinal])
    rundata.gaugedata.gauges.append([4, -94.13, 29.58, rundata.clawdata.t0, rundata.clawdata.tfinal])
    rundata.gaugedata.display_format = {1:"e10.2"}
    rundata.gaugedata.file_format = 'ascii'
    rundata.gaugedata.q_out_fields = {1:[True, True], 
                                      2:[False, True, False]}
    rundata.gaugedata.aux_out_fields = {1:[False, False, False, False, True, True, True], 
                                        2:[False, False, False, False, False, False, True],
                                        3:[True]}
```

rjleveque:
I like the first version better, it seems clearer, and this get converted into gauges.data for the Fortran code to read, so it seems like either version could make the same file for Fortran?

It seems like we could also allow things like:

```
    rundata.gaugedata.q_out_fields = {1:[1, 2], 
                                      2:"all"}
```
or even 
```
    rundata.gaugedata.q_out_fields = {'all': 'none'}
```
to specify that no q fields should be printed at any gauge.

I suggest that we also add something like:
```
    rundata.gaugedata.min_time_increment = 10.
```
to indicate that you only need the gauge data every 10 seconds.  This would help reduce the size of the output when one has many gauges in a region where the time step is tiny (as often happens onshore in a tsunami model).  Then we would have to keep track of the last time a value was recorded at each gauge and add a new value to the buffer only if the current time is at least this far beyond the last time.

I suppose one might want different time increments for different gauges, which might argue for this value to be added to each list in `rundata.gaugedata.gauges`, but a single value would be good enough for many purposes.

Regarding outputting `eta` in GeoClaw, if it's easy to output both `q(1)=h` and `aux(1)=B` then the sum of these gives `eta` so we might not need it.  Then new gauge output wouldn't be compatible with old plotting scripts of course.  Also a possible danger is that in deep ocean the first 4 digits of `q(1)` and `aux(1)` cancel out in computing `eta`, so unless sufficient digits are printed it's impossible to recover `eta` accurately.  This is the reason we decided to print `h` and `eta` rather than `h` and `B` originally.

mandli:
I am working on a new branch now that incorporates suggestions from @rjleveque including the `min_time_increment`.  The one slight difference from above is that specifying a parameter meant for all gauges is implemented
```python
rundata.gaugedata.q_out_fields = 'none'
```
rather than
```python
rundata.gaugedata.q_out_fields = {'all': 'none'}
```
If there's preference for the latter due to readability let me know, it's not hard to change.

I will close this PR once the replacement is ready for actually being merged.

rjleveque:
I like your suggested change, seeing it again,
```
   rundata.gaugedata.q_out_fields = {'all': 'none'}
```
is confusing!

Can you also specify, e.g.
```
   rundata.gaugedata.q_out_fields = [1,2]
```
if you want all gauges to output fields 1,2, rather than having to list every gauge?

mandli:
Yes, that's how it's implemented right now.  I figured this might be the most common usage so made sure to implement it this way.

mandli:
This has been replaced by #232.

PR #220, opened 2016-08-09
WIP:  Enable Storms from Atmospheric Data

mandli:
Enables explicit specification of storm fields from an outside source.


mandli:
This is based on work by @MarcKjerland to integrate ASCII WRF input.

PR #228, opened 2016-10-02
Update Regression Data for Changes to Riemann Solver

mandli:
This PR updates the regression test data due to changes in the GeoClaw Riemann solver (clawpack/riemann#111).  It also adds the git status and diffs of the currently checked out repositories.


rjleveque:
Out of date changes to regression test data.

PR #232, opened 2016-11-24
Add additional gauge options

mandli:
This PR adds a number of additional gauge output options as described in clawpack/amrclaw#178.  Note that $\eta$ is still output with the gauge data always in the same location (at the end of the q fields).  This also replaces #219.


mandli:
Other relevant PRs:
 - amrclaw#178
 - clawutil#106
 - classic#75
 - clawpack#94

rjleveque:
Works for me.

PR #236, opened 2016-12-21
Regression data update for clawpack/riemann#111

rjleveque:
I reran the regression tests using the new GeoClaw Riemann solver merged in clawpack/riemann#111.

Some issues:

In tests/bowl_slosh I got the following error and so I got rid of the relative import:

```
$ python regression_tests.py True
  File "regression_tests.py", line 70, in <module>
    test.setUp()
  File "regression_tests.py", line 40, in setUp
    from .make_fgmax_grid import make_fgmax_grid1
ValueError: Attempted relative import in non-package
```
@mandli, was there a reason for this relative import?

I also removed the `claw_git_diff.txt` files from the `regression_data` directories, because these files get filled up with a recursive bunch of garbage when they are committed to git and git does a `git diff` on the old version to make a new version.

I suggest in the future we only commit the `claw_git_status.txt` files, and in general the working directories should be clean when we do this so there won't be any uncommitted changes except in the `regression_data` directory itself.


mandli:
> @mandli, was there a reason for this relative import?

Nope, was probably being lazy.

> I also removed the claw_git_diff.txt files from the regression_data directories, because these files get > filled up with a recursive bunch of garbage when they are committed to git and git does a git diff on > the old version to make a new version.
> I suggest in the future we only commit the claw_git_status.txt files, and in general the working directories should be clean when we do this so there won't be any uncommitted changes except in the regression_data directory itself.

I agree that it does not make sense to keep the diffs.  They can always be reconstructed via the commit hash anyway.

rjleveque:
Actually the diff's can't be reconstructed -- they show changes in the working directories that have not been committed.  But the point is that we should check to see that there aren't any such changes and resolve them (or at least make sure they don't affect the tests being run) before creating new regression data.  Once we've done that, we don't need to archive the diff file.

We should probably also have a policy of rerunning the tests and recreating and committing the `claw_git_status.txt` files once more before each release, so that the commits they point to are pretty much up to date with the code that's released, both to reconfirm that all the tests work for the release and for clarity to someone trying to figure out what version the tests worked on.

Note: Travis tests for this PR won't pass until we update clawpack/clawpack with the latest riemann commits.

rjleveque:
@mandli I think you used the relative import because otherwise it doesn't work on Travis, with the error below.  I just tried moving the import statement up to the top of the file, thinking maybe it was in the output directory when the class was invoked, but that didn't help.  

There are a lot of confusing discussions on relative imports on stackoverflow, which didn't help much.  I don't see what was wrong with the initial version since there is an `__init__.py` file in this directory.

@ketch, maybe you can shed light on this?

```
ERROR: Test bowl-slosh example
  File "/home/travis/build/clawpack/geoclaw/tests/bowl_slosh/regression_tests.py", line 40, in setUp
    from make_fgmax_grid import make_fgmax_grid1
ImportError: No module named make_fgmax_grid
```

ketch:
One alternative is to use the full absolute import (though it feels ugly to me):

    from clawpack.geoclaw.tests.bowl_slosh.make_fgmax_grid import make_fgmax_grid1

I also think this will work:

    from . import make_fgmax_grid

(and then call `make_fgmax_grid.make_fgmax_grid1()`).  Sorry, I don't have time to test write now; have to go wrap my presents...

rjleveque:
@mandli: do you understand the error that the Python3 build now gives?

```
  File "/home/travis/virtualenv/python3.5.2/lib/python3.5/site-packages/numpy/lib/_datasource.py", line 215, in __del__
    if self._istmpdest:
AttributeError: 'DataSource' object has no attribute '_istmpdest'
```

mandli:
No idea.  Do you have any idea where it's being raised?

rjleveque:
Not except what's in the Travis log.

ketch:
I saw this error message once recently but can't remember what turned out to be the issue.  Will take a look later today.

ketch:
I'm a bit confused.  As far as I can tell, the exception quoted by @rjleveque above is ignored; there are other exceptions that cause test failure, but they seem to be caused by building against old versions of other parts of clawpack.  The first one is

```
ERROR: Test bowl-slosh example
  File "/home/travis/build/clawpack/geoclaw/tests/bowl_slosh/regression_tests.py", line 53, in runTest
    self.write_rundata_objects()
  File "/home/travis/build/clawpack/geoclaw/clawpack/clawpack/clawutil/test.py", line 214, in write_rundata_objects
    self.rundata.write()
  File "/home/travis/build/clawpack/geoclaw/clawpack/clawpack/clawutil/data.py", line 505, in write
    data_object.write()
  File "/home/travis/build/clawpack/geoclaw/clawpack/clawpack/amrclaw/data.py", line 90, in write
    self.data_write(file, self.aux_type, 'aux_type')
NameError: name 'file' is not defined
```

But in master, line 90 of amrclaw/data.py is

    self.data_write('aux_type')

Looking at `geoclaw/.travis.yml`, I don't see why you're getting an old version of amrclaw...

ketch:
Okay, it's clear now: the line in question was fixed in https://github.com/clawpack/amrclaw/commit/b4f8b4e8c5fe72fd8cf4220ea2b2dfa67847a5dc, but clawpack/clawpack is pointing to a slightly older version of AMRClaw.  So I think updating clawpack/clawpack to a newer version of AMRClaw will get rid of some or all of the errors we're seeing here.

PR #239, opened 2017-01-05
change Makefile.geoclaw to use fgmax_interpolate0.f90

rjleveque:
As suggested in documentation,
  http://www.clawpack.org/fgmax.html
(which needs to be updated for 5.4.0 Makefile).


rjleveque:
Switching to `fgmax_interpolate0.f90` caused one of the tests to fail due to small changes in the fgmax output:

```
File "/Users/rjl/git/clawpack/clawpack/geoclaw/test.py", line 84, in check_fgmax

AssertionError: 
 data: [  20.33055828  266.74951035], 
 expected: [  20.28603161  266.98593275]
```

I modified the Makefile to switch back to `fgmax_interpolate0.f90` for this example.

PR #242, opened 2017-02-11
Move Pressure Gradient Calculation to Riemann Solver

mandli:
In order to preserve well-balancing properties for strong pressure gradients the contribution from the contribution is now incorporated into the Riemann solver itself.  The PR in the Riemann repository is clawpack/riemann#116.

### DIFF
--- a/examples/multi-layer/plane_wave/qinit.f90
+++ b/examples/multi-layer/plane_wave/qinit.f90
@@ -1,7 +1,8 @@
 subroutine qinit(meqn,mbc,mx,my,xlower,ylower,dx,dy,q,maux,aux)
     
+    use geoclaw_module, only: rho
     use qinit_module, only: qinit_type, add_perturbation
-    use multilayer_module, only: num_layers, rho, eta_init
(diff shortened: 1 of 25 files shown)


mandli:
I moved `rho` into the `geoclaw_module` along with a couple of other, similar parameters.  There is an analogous change to the riemann solvers.

I also removed the parameter in the multilayer code that stored the density in place of the new `geoclaw_module` parameter.  The only users that will actually be effected are those using the multilayer code.

rjleveque:
@mandli: I just reran Travis and it still fails but maybe because it needs clawpack/riemann#116.  With that the nosetests pass on my laptop, after fixing another integer divide issue in python3, in `tests/old_topotools.py` at line 134.  I don't know why this didn't cause other PRs to fail, but it also seems somewhat random on my laptop whether this causes a problem or not with python3.

If you think this is ready to merge, we should first merge in clawpack/riemann#116 and update clawpack/clawpack and then see if the tests pass here.

mandli:
Just did a PR to clawpack/clawpack.

PR #243, opened 2017-02-12
Update NetCDF Topography Test Data

mandli:
I don't think this was updated when we were changing all the gauge IO stuff and reran the tests due to changes to the Riemann solver.  This updates these tests (and won't impact Travis unfortunately).  Below is a comparison between the old and new gauge data.

gauge_comparison


mandli:
This was incorporated into #249

PR #245, opened 2017-02-22
Setaux debug

aks2203:
The boundary error has been fixed with some corrections to setaux.f and stepgrid.f90. Corrections first made on mjberger's branch multilayer-fix.


aks2203:
Multi-layer plane wave example working now with flat bathymetry.

PR #249, opened 2017-04-05
Fix NetCDF Topography Reading

mandli:
This modifies the NetCDF topography reading so that it does not assume as much about the structure.


mandli:
This needs to be tested thoroughly as Travis does not run the NetCDF test.

PR #256, opened 2017-05-03
add INTERPOLATE parameter to gauges_module.f90

rjleveque:
Setting to .false. will suppress bilinear interpolation and always use value from cell containing gauge.  It is set to .true. so this shouldn't change the behavior from the previous version.  

Note that the interpolating version uses bilinear interp if all surrounding cells are wet, but takes the cell value if at least one is dry. 

This modification is similar to the `fgmax_interpolate0.f90` routine that takes the cell value rather than interpolating.  

Eventually we might want to add options to setrun.py to select whether to interpolate or not for both of these routines.

### DIFF
--- a/src/2d/shallow/gauges_module.f90
+++ b/src/2d/shallow/gauges_module.f90
@@ -84,6 +84,8 @@ module gauges_module
     integer, allocatable, dimension(:) ::  mbestsrc, mbestorder, &
                           igauge, mbestg1, mbestg2
 
+    logical, parameter :: INTERPOLATE = .true.
+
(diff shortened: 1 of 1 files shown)


mandli:
We could add it like the other new gauge options.  I could maybe take a crack at it if you would like.

rjleveque:
Let's wait and redo the way fgmax is handled at the same time.  I first want to experiment more with this gauge approach vs. interpolation.  We  should also consider whether to allow a similar option more generally for gauges in amrclaw and pyclaw.

PR #257, opened 2017-05-06
remove unnecessary warning message

rjleveque:
After checking again, I think `fgmax_interpolate0.f90` is correct and I have removed the warning message that raised concerns in https://groups.google.com/forum/#!topic/claw-users/r_eeWVx6tgw


rjleveque:
@mandli, @ketch: I am confused about why I needed to change `/` to `//` for integer division in `topotools` and `dtopotools` in order to get the Python 3 tests to pass.  Actually that seems right, but what confuses me is why earlier PRs such as #256 passed without this change??

mandli:
I am not sure myself but given that this is a floating point issue it may have behaved slightly differently for some other reason.  Another possibility is that the stochasticity of travis struck again.

PR #259, opened 2017-05-10
Update holland_storm_module.f90

lidiabressan:
corrected days of the month


mandli:
This is failing because of another commit that was just made that has not made it completely upstream yet @lidiabressan.  Hold on a sec while we work that out.

PR #271, opened 2017-06-16
Basic refactor of multilayer plotting

mandli:
Refactor and fix multilayer plotting.

### DIFF
--- a/examples/multi-layer/plane_wave/setplot.py
+++ b/examples/multi-layer/plane_wave/setplot.py
@@ -253,9 +253,11 @@ def setplot(plotdata=None, bathy_location=0.15, bathy_angle=0.0,
                                          figno=301)
     plotfigure.show = True
     plotfigure.clf_each_gauge = True
+    plotfigure.kwargs = {'figsize': (14, 4)}
 
     # Set up for axes in this figure:
     plotaxes = plotfigure.new_plotaxes()
+    plotaxes.axescmd = 'subplot(1, 2, 1)'
     plotaxes.xlimits = [0.0, 1.0]
     plotaxes.ylimits = top_surface_limits
     plotaxes.title = 'Top Surface'
@@ -265,17 +267,11 @@ def setplot(plotdata=None, bathy_location=0.15, bathy_angle=0.0,
     plotitem.plot_var = 6
     plotitem.plotstyle = 'b-'
 
-    # Bottom
-    plotfigure = plotdata.new_plotfigure(name='Bottom Surface Gauge',
-                                         type='each_gauge',
-                                         figno=302)
-    plotfigure.show = True
-    plotfigure.clf_each_gauge = True
-
     # Set up for axes in this figure:
     plotaxes = plotfigure.new_plotaxes()
+    plotaxes.axescmd = 'subplot(1, 2, 2)'
     plotaxes.xlimits = [0.0, 1.0]
-    # plotaxes.ylimits = internal_surface_limits
+    plotaxes.ylimits = internal_surface_limits
     plotaxes.title = 'Bottom Surface'
 
     # Plot surface as blue curve:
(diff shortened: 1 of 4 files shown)


mandli:
@aks2203 can you help with this by chance?  I have done only a bit of refactoring but it would be good to incorporate your additions while we are at this if possible.

rjleveque:
I guess this isn't done yet, but I notice the surface plots aren't showing anything changing.

Also, something strange happens in making the movie for this figure: several gauge plots are included.  The html file `file://.../clawpack/geoclaw/examples/multi-layer/plane_wave/_plots/frame_allframesfig1001.html` includes the lines below.  I've never seen this before and haven't yet investigated beyone running `make plots` with and without plotdata.parallel set and get the same results either way.

```
frames = ['frame0000fig1001.png', 'frame0001fig1001.png', ... 'frame0040fig1001.png', 
'gauge0000fig1001.png', 'gauge0001fig1001.png', 
'gauge0002fig1001.png', 'gauge0003fig1001.png', 'gauge0004fig1001.png']
```

mandli:
The first does not work for me either and I am not sure yet why (working on that).  I had not noticed the latter though, will look into a bit but I was wondering if it might be due to the non-assigned figure numbers?

mandli:
Also, aks2203 and I have been discussing what to do about colorbars for the internal layers.  Right now we were going to try and mimic what's done for the top surface but add a second colorbar as the depth and surface colorbars are centered in different places.  Any other ideas?

rjleveque:
Yes, adding `figno=` to each new figure fixes the gauge issue.

Regarding colorbars, it looks like the plots for the top surface and internal layer are on two different axes with separate colorbars already, so I'm not sure what the issue is?

mandli:
> Yes, adding `figno=` to each new figure fixes the gauge issue.

Is this an issue that's easy to fix or should we always assign a `figno` to a gauge.

> Regarding colorbars, it looks like the plots for the top surface and internal layer are on two different axes with separate colorbars already, so I'm not sure what the issue is?

It's for when there's wetting and drying on the internal layer.  Switching between the surface and depth in that case is a bit weird as the depth should be strictly positive and surface is going to arbitrarily different from the depth values.

rjleveque:
I think every other setplot explicitly sets figno, usually to 300, 301, etc. for gauges.  Maybe we should allow not setting it, but for now easiest to just set them?

mandli:
I think I fixed what needed fixing although I kind of would like to establish the way that inundation is plotted if it does not take too long.

rjleveque:
@mandli: Are you suggesting we not merge this yet?  Is the fix to inundation plotting something you want in 5.4.1?

aks2203:
I have some plots I made with multilayer inundation. These first set have the inundated area color bar set from white to red ({1.0:'r',0.0:'w'}), and the cross section plots only showing eta, with different colors for each surface. The cross sectional plots help depict where the overhead views have trouble, as one can see where the cells go from deep to shallow and the surface looks more elevated.

frame0000fig4
frame0001fig0
frame0009fig0
frame0010fig4

Here the inundated area has color bar {1.0:'r',0.0:'c'}, but with both of these color bars there is discontinuity where the initially wet surface meets the inundated cells.
frame0001fig0-1
frame0009fig0-1

This third set is an attempt to fix the discontinuity in color between where the surface is being plotted and the depths of the inundated cells are being plotted, but it doesn't quite look the way I'd hoped.
frame0001fig0-2
frame0009fig0-2

So I don't have a good solution, but perhaps looking at these will be helpful.

mandli:
@rjleveque it's not absolutely necessary but it would be nice to come up with a convention.  Given that the examples do not need this yet maybe it's not essential?  If we do we should move the discussion to an issue then.

rjleveque:
@mandli: The code in this PR seems to break plotting in `examples/storm-surge/ike`, where I get figure like what's below (in both Python 2 and 3).
ike

mandli:
Huh, that's weird.  I will have to look into it.

mandli:
Oh, I think it should work now.  This PR was branched from clawpack/master before the storm surge fix was merged.  I rebased on the current clawpack/master so it should work now again.

PR #272, opened 2017-06-21
Chile2010 titles

rjleveque:
Improve the title on plots to use hr:min:sec format.

### DIFF
--- a/examples/tsunami/bowl-slosh/setplot.py
+++ b/examples/tsunami/bowl-slosh/setplot.py
@@ -101,10 +101,10 @@ def setplot(plotdata=None):
     plotaxes.ylimits = [-0.15,0.3]
     plotaxes.title = 'Cross section at y=0'
     def plot_topo_xsec(current_data):
-        from pylab import plot, hold, cos,sin,where,legend,nan
+        from pylab import plot, cos,sin,where,legend,nan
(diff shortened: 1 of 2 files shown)


mandli:
Wouldn't it be easier to use `datetime`?

rjleveque:
Do you know how to easily?  I guess you can make a `datetime.time` object and then use `time.isoformat()` to get such a string, but to make the object in the first place you have to convert from seconds to hours, minutes, seconds (since it requires seconds <= 60 ) so that all seems more complicated than just constructing the string as I did.

mandli:
Had to remember how I did this once as it is not exactly the most intuitive thing.  If you define
```
my_time = datetime.datetime(2017, 1, 1) + datetime.timdelta(seconds=my_seconds)
```
the object `my_time` will be an offset from the base date with `my_seconds > 60` allowed.  Then you can use the string formatting of `datetime` to output the string you are looking for.  I think
```
my_time.strftime("%X")
```
might do what you want.

rjleveque:
This PR is obsolete - as of v5.9.1 a `ClawPlotAxes` object can have a `title` attribute that contains `h:m:s` or `d:h:m:s`.

See https://www.clawpack.org/dev/ClawPlotAxes.html

PR #275, opened 2017-06-27
Add back default land bounds

mandli:
In some instances the land bounds were not being set making the land plotting janky (technical term for awful).

### DIFF
--- a/examples/storm-surge/ike/setplot.py
+++ b/examples/storm-surge/ike/setplot.py
@@ -59,21 +59,11 @@ def setplot(plotdata=None):
                                   kwargs={"markersize": 4})
 
     # Color limits
-    surface_limits = [physics.sea_level - 5.0, physics.sea_level + 5.0]
-    surface_ticks = [-5, -4, -3, -2, -1, 0, 1, 2, 3, 4, 5]
-    surface_labels = [str(value) for value in surface_ticks]
+    surface_limits = [-5.0, 5.0]
     speed_limits = [0.0, 3.0]
-    speed_ticks = [0, 1, 2, 3]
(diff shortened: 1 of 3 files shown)


mandli:
Something is wrong with the colorbars that I need to fix (they are off center).  Looking into it.

mandli:
@rjleveque are you seeing this in any other examples?  I am somewhat suspicious that this might be related to the `pobj` issue (https://github.com/clawpack/visclaw/pull/219).  Note that I am seeing this in both 2.7 and 3.x but for matplotlib version 2.0.x.

figure_1003

mandli:
One thing to note here is that in this particular example I *try* to set the colorbar ticks manually.  Not doing this does not fix the problem but makes it mildly better.

rjleveque:
@mandli: Do you mean the fact that colorbar colors aren't centered about 0?  But it looks like the ocean is white indicating no disturbance at some negative elevation.  Maybe you have `sea_level` set to something other than 0 and the setplot is set to center around that value (I do this for some situations).

mandli:
I think I figured out the problem.  For the Ike example `sea_level != 0.0` but in matplotlib version < 2 I translated the colorbar so that it looked like `sea_level` was the zero point but plotted as zero.  It appears that the colorbar now always takes the `vmin` and `vmax` from the `pobj` rather than respecting `clim`.  I am not sure if this is completely true but testing seems to indicate this to be the case.  

So it's not probably related to clawpack/visclaw#219 but another fix for the plotting needs to be made.  New commit forthcoming.

mandli:
This is now working in Python2 but Python3 has the problem in clawpack/visclaw#219 as described.  The fix there seems to work for Python2 and Python3.

PR #278, opened 2017-09-26
Add GeoTIFF to Topography Types

mandli:
Adds ability to read in GeoTIFF files into the `Topography` class.  Does not yet support writing out these files.

### DIFF
--- a/src/python/geoclaw/topotools.py
+++ b/src/python/geoclaw/topotools.py
@@ -517,6 +517,9 @@ class Topography(object):
                         self.read(mask=mask)
                     # Generate arrays
                     self._X, self._Y = numpy.meshgrid(self._x, self._y)
+                elif abs(self.topo_type) in [4, 5]:
+                    # Generate arrays
+                    self._X, self._Y = numpy.meshgrid(self._x, self._y)
                 else:
                     raise ValueError("Unrecognized topo_type: %s" % self.topo_type)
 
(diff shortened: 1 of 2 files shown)


rjleveque:
is there a simple example with some test data that could be included somewhere?

We probably don't want to put this in `tests` since in my experience installing `gdal` can be nontrivial.  

See also #279.

donnaaboise:
It is relatively painless to install GDAL through conda.     I have some instructions on the Google Earth page [here](http://www.clawpack.org/googleearth_plotting.htmll).   The key is making sure that the GDAL_DATA environment variable points to the correct shared data files - e.g. anaconda/share/gdal. 

A recent GDAL bug (which may be fixed by now) is that you need to update the 'xy' library - also from conda.  Apparently, this dependency didn't make it into the dependency list.

mandli:
Installing via homebrew on OS X seems to have worked but I agree that a test should be skipped if it cannot find `gdal`.  The NetCDF reading already does this so I will just copy from there.  Documentation should also follow along with this.  @donnaaboise what documentation on installing gdal did you put in the docs for the google earth stuff?

donnaaboise:
I included the following two lines : 

conda install gdal
export GDAL_DATA=$ANACONDA/share/gdal

and then a simple test (which involved downloading a sample python script and a figure file to test the gdal tiling routines). 

I suppose somebody should test this out, though, as I have not done this install recently  (although I have used GDAL when plotting for Google Earth, and my installation seems to work). 


-----------------------------------
Donna Calhoun
Associate Professor
Department of Mathematics, MG241A
Boise State University
1910 University Drive
Boise, ID 83725-1555
(208) 426-3386
http://math.boisestate.edu/~calhoun
-----------------------------------

mandli:
You did not need to add the fix for `xy`?

donnaaboise:
I check the link on the page referencing that bug and that issue has been fixed.  See

https://github.com/ContinuumIO/anaconda-issues/issues/951 <https://github.com/ContinuumIO/anaconda-issues/issues/951>

(I should update the Google Earth vis. page on the Clawpack docs!)



-----------------------------------
Donna Calhoun
Associate Professor
Department of Mathematics, MG241A
Boise State University
1910 University Drive
Boise, ID 83725-1555
(208) 426-3386
http://math.boisestate.edu/~calhoun
-----------------------------------

rjleveque:
For me `conda install gdal` seems to have worked by itself, once I un-installed the brew version.  Maybe instead I could have used `GDAL_DATA` to point properly to the brew version.

donnaaboise:
Yes - the GDAL “data” (projection files, etc) needs to be compatible with the installed version of GDAL.  

-----------------------------------
Donna Calhoun
Associate Professor
Department of Mathematics, MG241A
Boise State University
1910 University Drive
Boise, ID 83725-1555
(208) 426-3386
http://math.boisestate.edu/~calhoun
-----------------------------------

mandli:
`GDAL_DATA` definitely needs to be set for the `brew` version to work.

mandli:
Would be good to come back to this feature soon.  Adding “good first issue” to this.

PR #277, opened 2017-09-19
Multilayer amr

aks2203:

PR #280, opened 2017-11-06
WIP: Output cumulative CPU timing for each frame

rjleveque:
I added code to `valout.f` to output the cumulative CPU and wall time on each level to each `fort.t` file, to make plots like the one shown here, produced by the sample code in `src/python/geoclaw/plot_timing_stats.py` after running the `examples/tsunami/bowl-radial` code.

This PR is primarily to start a discussion about whether the `fort.t` file is a good place to put this and whether other timing data should also be output.

I've found this useful for analyzing some long GeoClaw runs and @aks2203 plans to start using this for some timings of the new multilayer code in #277.

The same changes could be made to `valout.f` in the amrclaw code, of course.  General tools for plotting the results should ultimately go in amrclaw or clawutil perhaps, not in geoclaw.

cputime

### DIFF
--- a/src/2d/shallow/valout.f
+++ b/src/2d/shallow/valout.f
@@ -22,6 +22,7 @@ c     # set outaux = .true. to also output the aux arrays to fort.a<iframe>
       real(kind=8), allocatable :: qeta(:)
 
       integer :: clock_start, clock_finish, clock_rate
+      integer :: timeTick_sofar, tick_clock_sofar
 
       iadd(ivar,i,j)  = loc + ivar - 1 + nvar*((j-1)*mitot+i-1)
       iaddaux(iaux,i,j) = locaux + iaux-1 + naux*(i-1) +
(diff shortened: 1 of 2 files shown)


mandli:
A couple of thoughts:
1. I like the plots and this would be nicer than having to save the output to the terminal
1. I was wondering maybe if we would rather just output another file rather than change the format of `fort.t`
1. It would be nice if we provided easy tools for also extracting AMR stats, something we could all package into something called "analysis" maybe?
1. Can we merge whatever gets decided here into https://github.com/clawpack/amrclaw/pull/202

mandli:
@rjleveque do you think this is ready to merge in by chance?

mandli:
clawpack/amrclaw#202 now contains output similar to what was being recorded in this PR but now in a separate `timing.csv` file.  Once we figure out a place to put the script included in this PR we can close this and merge clawpack/amrclaw#202.

rjleveque:
A newer version of this is now in clawpack/amrclaw#202.

The modified `valout.f90` code has not yet been ported to geoclaw.

PR #283, opened 2018-01-11
added debugging info to upbnd

mjberger:
can't reproduce the bug, but added debugging info in case it happens again


mandli:
I was thinking maybe we could simply get people to use this without merging this in to master but it probably would not hurt either way.

mjberger:
why? is there a problem with merging it in?

mandli:
Mostly I was thinking that both of I the current instances are already on branches and this might confuse things given that this is strictly debugging info.  On the flip side if someone finds a new case of this they would already have the debugging code if they were up to date.

PR #285, opened 2018-01-22
Add New Storm Formats and Capabilities

mandli:
This is a significant update to the storm surge capabilities in GeoClaw and includes:
 - Now distinguishes "model" storms from "data driven" storms.  The former uses a specified wind model (which now support for other fields has increased) and the latter as of yet unimplemented capability for input from say another package to specify and interpolate wind and pressures fields directly.
 - A new `storm` module that reads in multiple different types of storm tracks and intensities and output them into a unified  and simplified input type specific to GeoClaw
 - Adds a number of utilities for dealing with storm data
 - Example and tests now directly download storm data from the best-track database

Contributors include @hudaquresh and @sarasamuel.


mandli:
Note that the tests are locally failing for me still in the multilayer test on Python 3.  I think this may not be due to this code but the verdict is still out on that.

mandli:
Just as an update to this PR.  I think we have traced the issue with the failed multilayer tests to a previous bug with setting the manning's N friction coefficient that now is fixed in this PR.  The regression data has been updated accordingly.

mandli:
This is now failing because of an import issue on Travis.  Will look into it.

mandli:
It looks like the remaining errors are due to the issue with amrclaw.

mandli:
Ug, those are the worst, attempting to fix.

rjleveque:
@mandli: Tests are now passing, is this ready to merge?

mandli:
Yeah, I think so.

src/python/geoclaw/surge/storm.py
```diff
@@ -649,60 +660,81 @@ def write_geoclaw(self, path, verbose=False, max_wind_radius_fill=None,
...
-                num_casts -= 1
+        # Create list for output
+        # Leave this first line blank as we need to count the actual valid lines
+        # that will be left in the file below
+        num_casts = 0
+        data_string = [""]
+        if self.time_offset is None:
+            # Use the first time in sequence if not provided
+            self.time_offset = self.t[0]
+        import pdb; pdb.set_trace()
```
rjleveque: This pdb statement is causing the travis failure.

mandli: Thanks for catching that.

mandli:
@rjleveque I added the changes to bc2amr we discussed.  Given that I was changing a significant portion of the code I also refactored it to F90.  I can do something similar to AMRClaw if it looks ok.

src/2d/shallow/bc2amr.f90
```diff
@@ -0,0 +1,280 @@
...
+    !-------------------------------------------------------
+    ! Top boundary:
+    !-------------------------------------------------------
+    if (yhi_patch > yupper + hymarg) then
+
+        ! number of grid cells lying outside physical domain:
+        nyt = int((yhi_patch - yupper + hymarg) / hy)
+        jbeg = max(ncol - nyt + 1, 1)
+
+        select case(mthbc(3))
```
rjleveque: I think this should check `mthbc(4)` not 3.

src/2d/shallow/bc2amr.f90
```diff
@@ -0,0 +1,280 @@
...
+                    end do
+                end do
+                ! negate the normal velocity:
+                do j = 1, ncol
+                    do i=1, nxl
+                        val(2, i, j) = -val(2, i, j)
+                    end do
+                end do
+
+            case default
```
rjleveque: We should also allow `case(4)` (sphere BCs) with a `continue` like `case(2)`, both here and for the other three boundaries.  

In principle we never get here in this case, since the module variable `spheredom` is checked earlier and should jump out of this routine, but might add clarity.

The spherical BCs haven't been tested in a while as far as I know, not sure if they still work properly...

src/2d/shallow/bc2amr.f90
mandli: Yeah, that would probably be good wouldn't it?

src/2d/shallow/bc2amr.f90
mandli: Added.

rjleveque:
@mandli: I think you need to bring your changes to `valout.f` into the new `valout.f90` routine and delete `valout.f`.

mandli:
I think so.  Let me know if you see anything.

rjleveque:
Seems to be fine!

PR #287, opened 2018-03-08
Add function to fetch NOAA tide data

navravi:
**Description**

This PR adds a new function to the `util` package called `fetch_noaa_tide_data` that provides a convenient way of obtaining NOAA tide station data given a station ID and begin and end dates/times. The data is retrieved using the [NOAA CO-OPS API for Data Retrieval](https://tidesandcurrents.noaa.gov/api/) and cached locally in the scratch directory by default.

**Testing**

Four unit tests were added to `tests/test_util.py` which test all code paths through the new function except for the first `if` block that handles the case when `cache_dir` is not specified. A manual test was performed to verify that the scratch directory is used for caching when `cache_dir` is `None`.
```
$ nosetests tests/test_util.py 
....
Ran 4 tests in 0.006s

OK
```


src/python/geoclaw/util.py
```diff
@@ -141,3 +150,121 @@ def inv_haversine(d,x1,y1,y2,Rsphere=Rearth,units='degrees'):
...
+                         datum='STND', units='metric', cache_dir='_cache'):
+    r"""Fetch water levels and tide predictions at given NOAA tide station.
+
+    The data is returned in 6 minute intervals between the specified begin and
+    end dates/times.  A complete specification of the NOAA CO-OPS API for Data
+    Retrieval used to fetch the data can be found at:
+
+        https://tidesandcurrents.noaa.gov/api/
+
+    By default, retrieved data is cached in the subdirectory '_cache' of the
```
mandli: I am not sure we want to enable another way of caching local files as we already have a `scratch` directory in `$CLAW/geoclaw`.  I actually would not be opposed to using this instead except that sometimes multiple examples will use the same files which would mean redundant data download.  Given this I would probably suggest sticking these in the `scratch` directory instead.

src/python/geoclaw/util.py
```diff
@@ -141,3 +150,121 @@ def inv_haversine(d,x1,y1,y2,Rsphere=Rearth,units='degrees'):
...
+        cache_path = get_cache_path(product)
+
+        # use cached data if available
+        if os.path.exists(cache_path):
+            print('Using cached {} data for station {}'.format(
+                product, station))
+            return parse(cache_path, col_idx, col_types, header=True)
+
+        # otherwise, retrieve data from NOAA and cache it
+        print('Fetching {} data from NOAA for station {}'.format(
```
mandli: Might suggest allowing this to be suppressed by adding a `verbose` keyword argument that defaults to `True`.

mandli:
Thanks @navravi.  

I made one comment regarding the `scratch`/`_cache` issues that we had already talked about.  A couple of ideas perhaps worth discussing a bit before merging:
 - Would it be useful to pass back a `GaugeSolution` instead of bar arrays?  I tend to think not as there would not be a good way of doing this with both the predicted and true signal and all we have is the water height.
 - Is there any other data that is fetch-able that might be useful?  One that comes to mind is the tidal constituents.  If there isn't a good way to fetch these it's probably fine to ignore.

src/python/geoclaw/util.py
navravi: Updated to use scratch directory for caching by default.

src/python/geoclaw/util.py
navravi: Sure, fixed.

navravi:
Thanks @mandli for the review. I've addressed the two inline comments in the latest diff.

If we were to pass back a `GaugeSolution` object, it seems that we wouldn't have data for many of the columns such as `level` and `q[1..3]`. Do you think it would be okay to leave those columns as `0`s (or some other default value)? Also, is there a way to use a `GaugeSolution` object without assigning it a `gauge_id` or using the corresponding `gauge<id>.txt` file? Let me know what you think makes the most sense.

It appears that NOAA does publish harmonic constituents (see [here](https://tidesandcurrents.noaa.gov/harcon.html?unit=0&timezone=0&id=8516945&name=Kings+Point&state=NY) for an example), but as far as I can tell they don't have an API to access the data. Of course, we could scrape their web page or use a library like [pytides](https://pypi.python.org/pypi/pytides/0.0.4) or [tappy](https://pypi.python.org/pypi/tappy) as a workaround. However, code that scrapes a website we don't control is probably a bad idea to put in a library since it could break at any time by a UI change by NOAA, and the two third party libraries haven't been updated since 2014/2015. (In fact, I couldn't even import `pytides` with Python3 last time I tried.) I'm definitely open to ideas on how to include this data if you feel that it would be important to have in geoclaw, or even implementing one of the hacky workarounds I mentioned as long as I'm not responsible when the code breaks. :wink:

Finally, regarding other fetchable data, let me know if anything listed [here](https://tidesandcurrents.noaa.gov/api/) under "Data Products" looks useful, and I'd be happy to include it.

mandli:
- I think return arrays is a better option for exactly the reasons you mentioned.
 - I think we should stick to data available through the API as I don't think it's critical information and I do not want to maintain such a thing.  It might be worthwhile mentioning this to someone at NOAA though.
 - @rjleveque might have some additional ideas on that.

PR #288, opened 2018-03-09
Add time scale to timespan routine for GE

donnaaboise:
Compatible with changes to VisClaw (plotpages.py)

### DIFF
--- a/src/python/geoclaw/kmltools.py
+++ b/src/python/geoclaw/kmltools.py
@@ -664,7 +664,7 @@ def kml_gauge(mapping):
 
 
 
-def kml_timespan(t1,t2,event_time=None,tz=None,tscale=3600):
+def kml_timespan(t1,t2,event_time=None,tz=None,tscale=1):
 
     r"""
     Create time strings necessary for sliders in Google Earth.  The time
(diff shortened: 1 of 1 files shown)


mandli:
Does this require https://github.com/clawpack/visclaw/pull/224?

donnaaboise:
Well, the `timespan` routine is called by `plotpages`, and is used by `plot2kml`.  So probably it should be with visclaw?  But where?  I didn't want to put it in plotpages, but am not sure where else it should go.  This routine figures out the correct time stamp to use for time frames in KML animations (viewed in Google Earth).

mandli:
Things like this are tricky to know where to put them.  I am not so concerned about that but I wanted to make a note that there is a cross-repository dependency here.  It looks like it won't be too big of a deal in any case.  I think we are also hoping to get a release done before the end of May.  The code looks fine to me so if you want to merge both of these I am fine with that.

donnaaboise:
Can you merge this in?  I don't have admin privileges

mandli:
Did you mean to close this?

donnaaboise:
ER ... yes ... but only because I thought we were done.  Can you merge in this PR?

PR #289, opened 2018-03-31
Googleearth : Add time span

donnaaboise:


donnaaboise:
(This replaces PR I accidentally closed)

PR #292, opened 2018-04-23
add timeTick and timeTickCPU back into amr2.f90 

rjleveque:
To fix deletion of these variables and print statements.  
Should now agree with amrclaw version again, and print out total time statistics at the end of a run.

The write statements were commented out in 5bdb1e2.  Not clear where the declaration of the variables got dropped, but it wasn't working at that time because of lack of proper `use` statements.


mandli:
I thought that this was removed because of something that the intel compiler was complaining about or am I thinking of something else that was related?

rjleveque:
I don't know.  @aks2203 told me he commented out the write statements because the module variables weren't included, but I couldn't figure out in what commit they got deleted.

If there's a problem with the Intel compiler on this then presumably there is with AMRClaw as well, which has the same write statements?

I think @mjberger cleaned up the timers at one point, and maybe remembers something about this?

Issue #298, opened 2018-05-05
where to flag regions in geoclaw?


rjleveque:
Currently in geoclaw, regions are checked twice.  We could eliminate call to `flagregions2` or have a local version that simply returns.

Regions are checked in `flag2refine2.f90` along with various other geoclaw-specific things.  In amrclaw we moved region checking to `flagregions2` so that if the user specifies Richardson flagging, or wants to provide their own `flag2refine2` with some other flagging criterion, this is decoupled from regions. Should we do the same in geoclaw?  In which case we should modify `flagregions2` to also check regions specifed by topo, dtopo, or qinit files (or deprecate these capabilities, since I think they often lead to confusion, and just do everything with explicit regions).

Usually Richardson isn't used with geoclaw, but @BrisaDavis has been experimenting with this in the context of the adjoint flagging methods, and the adjoint also requires a different [flag2refine2](https://github.com/BrisaDavis/geoclaw/blob/adjoint/src/2d_adjoint/flag2refine2.f90) routine (that currently is out of date, e.g. doesn't include storm module flagging stuff recently added by @mandli).  

We should discuss the best way to organize flagging so that it's easiest to modify in new ways in the future.

mandli:
I second this, flagging seems a bit more complex than it should be from an algorithmic standpoint.
 
 I did not realize that regions are checked twice though, isn't this due to checking the upper and  lower bounds?  Do you mean that a loop is cycled twice to check each rather than both at the same time?

rjleveque:
Both [geoclaw/src/2d/shallow/flag2refine2.f90](https://github.com/clawpack/geoclaw/blob/master/src/2d/shallow/flag2refine2.f90) and [amrclaw/src/2d/flagregions.f90](https://github.com/clawpack/amrclaw/blob/master/src/2d/flagregions.f90) loop over all cells.

mandli:
Oh I see what you mean now.  I was thinking of `allowflag` which also checks the region.

PR #299, opened 2018-05-05
flagregions.f90 never used, remove from Makefile.geoclaw

rjleveque:
`flagregions2.f90` is also in `Makefile.geoclaw` and is the one actually used.  `flagregions` is a left-over.

Bigger question to discuss: where should we flag regions in geoclaw, see #298.

### DIFF
--- a/src/2d/shallow/Makefile.geoclaw
+++ b/src/2d/shallow/Makefile.geoclaw
@@ -47,7 +47,6 @@ COMMON_SOURCES += \
   $(AMRLIB)/prepbigstep.f \
   $(AMRLIB)/bufnst2.f \
   $(AMRLIB)/flagger.f \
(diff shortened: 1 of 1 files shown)


mandli:
There are a few of these routines which I would prefer to remove unless there's a strong case to have either available at compile time.  I would almost suggest to remove `flagregions.f90` and move `flagregions2.f90` to `flagregions.f90`.

rjleveque:
Fine to remove `flagregions.f90`, but it's in amrclaw, not geoclaw.

There are many other routines with names `xxx2.f90` at this point  following some refactoring, so unless we want to clean up and rename them all and fix all the places they are called, I think we should leave this as `flagregions2.f90` for now.

mandli:
Good point, maybe something to raise more generally.

PR #302, opened 2018-05-09
fix to_lower function in utility_module.f90

rjleveque:
Looks like code was copied from to_upper and not properly modified?


rjleveque:
It should be checking if a character is in `upper_case` and if so replacing it with the corresponding character from `lower_case`.  That's what I think the new version is doing.  The old version looked some some things were copied from the `to_upper` function, which does the opposite,  and were not modified properly.

rjleveque:
@mandli: Good idea, I did so.

Issue #276, opened 2017-08-17
Proper interpretation of ArcInfo ASCII grid format?


rjleveque:
Resolve the issue raised here: https://groups.google.com/forum/#!topic/claw-users/a0sK7dwETvE

rjleveque:
The fix has been merged.

PR #303, opened 2018-05-09
Grid registration - handle xllcorner,yllcorner properly

rjleveque:
This is to deal with the issue that we always interpreted DEM files incorrectly if the header specifies `xllcorner, yllcorner`, see #276 and https://groups.google.com/forum/#!topic/claw-users/a0sK7dwETvE

This changes how the data is interpreted when DEMs of this form are read in to either the Fortran code (changes to `topo_module.f90`) or into Python using `topotools.py`, and therefore will change the computed results in these cases.

I am working on some documentation that explains all this better, comments welcome on the drafts at

 - http://depts.washington.edu/clawpack/sampledocs/v5.5.0alpha/topo.html
 - http://depts.washington.edu/clawpack/sampledocs/v5.5.0alpha/grid_registration.html

It would be good if someone experiments with this before we merge it.

Note: the chile2010 example uses a very old etopo file that specifies `xll,yll` in the header, which gets interpreted as would `xllcenter,yllcenter` in the new version, i.e. the same way it was interpreted in the old version.  So results for this test problem (and the unit tests in `geoclaw/tests`) do not change.  The chile2010 example should be updated and improved as a separate issue.

### DIFF
--- a/src/2d/shallow/topo_module.f90
+++ b/src/2d/shallow/topo_module.f90
@@ -723,6 +723,7 @@ contains
         character(len=10) :: x_dim_name, x_var_name, y_dim_name, y_var_name, z_var_name
         integer(kind=4) :: x_var_id, y_var_id, z_var_id
         logical :: verbose
+        logical :: xll_registered, yll_registered
         ! character(len=10) :: x_dim_name, y_dim_name, z_dim_name
         ! character(len=10) :: x_var_name, y_var_name, z_var_name
         ! integer :: ios, root_id, x_var_id, y_var_id, z_var_id, var_ids(10)
@@ -794,10 +795,14 @@ contains
                 read(iunit,'(a)') str
                 call parse_values(str, n, values)
                 xll = values(1)
+                xll_registered = ((index(str, 'xllcorner') > 0) .or. &
+                          (index(str, 'XLLCORNER') > 0))
 
                 read(iunit,'(a)') str
                 call parse_values(str, n, values)
                 yll = values(1)
+                yll_registered = ((index(str, 'yllcorner') > 0) .or. &
+                          (index(str, 'YLLCORNER') > 0))
 
                 read(iunit,'(a)') str
                 call parse_values(str, n, values)
@@ -811,6 +816,19 @@ contains
                 read(iunit,'(a)') str
                 call parse_values(str, n, values)
                 nodata_value = values(1)
+
+                if (xll_registered) then
+                    xll = xll + 0.5d0*dx
+                    write(6,*) '*** in file: ',trim(fname)
+                    write(6,*) '    Shifting xllcorner by 0.5*dx to cell center'
+                    endif 
+
+                if (yll_registered) then
+                    yll = yll + 0.5d0*dy
+                    write(6,*) '*** in file: ',trim(fname)
+                    write(6,*) '    Shifting yllcorner by 0.5*dy to cell center'
+                    endif 
+
 
                 xhi = xll + (mx-1)*dx
                 yhi = yll + (my-1)*dy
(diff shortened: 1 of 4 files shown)


mandli:
Raised an issue about updating the chile example.

src/2d/shallow/topo_module.f90
```diff
@@ -693,7 +693,7 @@ subroutine read_topo_header(fname,topo_type,mx,my,xll,yll,xhi,yhi,dx,dy)
 #endif
 
         use geoclaw_module
-        use utility_module, only: parse_values
+        use utility_module, only: parse_values, to_lower
```
mandli: Now I see why you caught the problem with `to_lower` :)

src/python/geoclaw/dtopotools.py
```diff
@@ -1134,7 +1134,8 @@ class SubFault(object):
       Each will be a tuple *(x, y, depth)*.
     
 
-    Top edge    Bottom edge
+    Top edge    Bottom edge ::
```
mandli: Does this render better in sphinx?

src/python/geoclaw/kmltools.py
```diff
@@ -676,7 +676,7 @@ def kml_timespan(t1,t2,event_time=None,tz=None,tscale=1):
     event_time : Start of event in UTC :  [Y,M,D,H,M,S], e.g. [2010,2,27,3,34,0]
     tz         : time zone offset to UTC.  e.g. +3 for Chile; -9 for Japan.
 
-    Time span element looks like :
+    Time span element looks like ::
```
mandli: Here with the colons again.

src/python/geoclaw/topotools.py
```diff
@@ -92,9 +92,7 @@ def create_topo_func(loc,verbose=False):
      - *loc* (list) - Create a topography file with the profile denoted by the
        tuples inside of loc.  A sample set of points are shown below.  Note 
        that the first value of the list is the x location and the second is 
-       the height of the topography.
-
-       **This figure doesn't show up properly in Sphinx docs...**
+       the height of the topography::
```
mandli: ...and colons here.

src/python/geoclaw/topotools.py
```diff
@@ -271,14 +269,20 @@ class Topography(object):
     A class representing a single topography file.
 
     :Properties:
+    
+    Note: Modified to check the `grid_registration` when reading or writing
```
mandli: This text should probably be put above the `:Properties:` header.

src/python/geoclaw/topotools.py
```diff
@@ -773,28 +800,50 @@ def read_header(self):
...
+                    # data points are offset by dx/2, dy/2
+                    self._x = x + dx/2.
+                    self._y = y + dy/2.
+                    print('*** Note: since grid registration is llcorner,')
+                    print('    will shift x,y values by (dx/2, dy/2) to cell centers')
+                else:
+                    # assume that x,y are cell center / data locations:
+                    self._x = x
+                    self._y = y
+                    print('*** Warning: Unrecognized grid_registration: %s' \
```
mandli: Might want to instead raise a warning so it could be caught in principle.  See [this stackoverflow post](https://stackoverflow.com/questions/15933741/how-do-i-catch-a-numpy-warning-like-its-an-exception-not-just-for-testing#15934081) for an example.

mandli:
Cursory tests using both intel and gcc with various flags all seems to pass.  Will keep on hitting this with tests though.

src/python/geoclaw/dtopotools.py
rjleveque: yes!

src/python/geoclaw/topotools.py
rjleveque: also works now in Sphinx

rjleveque:
@mandli confirmed it is ok to merge this in.

PR #308, opened 2018-05-16
added etopotools.etopo1_download_nc function

rjleveque:
to download subsampled subset of etopo1 directly from web using netCDF.

Requires `xarray` from https://xarray.pydata.org/en/stable/index.html

Sample code in the docstring.


src/python/geoclaw/etopotools.py
```diff
@@ -123,3 +124,95 @@ def etopo1_download(xlimits, ylimits, dx=0.0166666666667, dy=None, \
...
+        print('*** You need to install xarray, see:')
+        print('***     https://xarray.pydata.org/en/stable/')
+
+    x1, x2 = xlimits
+    y1, y2 = ylimits
+
+    assert (type(coarsen) is int) and (coarsen > 0), \
+           '*** coarsen must be a positive integer'
+
+    etopo1_url = 'https://www.ngdc.noaa.gov/thredds/dodsC/global/ETOPO1_Ice_g_gmt4.nc'
```
mandli: Maybe provide a way to over ride this?  I mostly thinking of temporary future proofing.

src/python/geoclaw/etopotools.py
```diff
@@ -123,3 +124,95 @@ def etopo1_download(xlimits, ylimits, dx=0.0166666666667, dy=None, \
...
+        # to save topofile for input to GeoClaw:
+        topo.write('etopo_sample_2min.tt3', topo_type=3, Z_format='%.0f')  
+
+    This should give a 2-minute resolution DEM of the Western Washington coast.
+    Note that etopo1 Z values are integers (vertical resolution is 1 meter)
+    and using `Z_format='%.0f'` will save as integers to minimize file size.
+    """
+
+    from clawpack.geoclaw import topotools
+    from pylab import find
```
mandli: Given that `pylab` is being threatened with deprecation can we use the direct `numpy` function instead?

src/python/geoclaw/etopotools.py
```diff
@@ -123,3 +124,95 @@ def etopo1_download(xlimits, ylimits, dx=0.0166666666667, dy=None, \
...
+    try:
```
mandli: I would say to just try and import `xarray` and let it raise an `ImportError`.  If we want to still print the message below maybe catch the error, print the message, and re-raise the exception.

src/python/geoclaw/etopotools.py
```diff
@@ -123,3 +124,95 @@ def etopo1_download(xlimits, ylimits, dx=0.0166666666667, dy=None, \
...
+    assert (type(coarsen) is int) and (coarsen > 0), \
+           '*** coarsen must be a positive integer'
+
+    etopo1_url = 'https://www.ngdc.noaa.gov/thredds/dodsC/global/ETOPO1_Ice_g_gmt4.nc'
+
+    if verbose:
+        print('Opening remote netCDF file')
+        print('    %s' % etopo1_url)
+
+    f = xarray.open_dataset(etopo1_url)
```
mandli: Does this need to be closed?  If so does xarray support the `with` syntax?

mandli:
A couple of question and comments:

1. Could we also support directly NetCDF support or is the xarray functionality required in this for loading the remote file?
1. I would suggest simply adding this functionality to the old function and providing a flag to use one or the other by providing some well-placed key word arguments.  That way we can continue to use the same function name and support both methods.

rjleveque:
Good suggestions @mandli, I'll keep working on this...

rjleveque:
@mandli: I removed the new function from etopotools and created a better version in topotools that can also be used to read other netCDF files. 

It now uses `netCDF4` unless the parameter `return_xarray` is `True`, in which case it uses `xarray`.  The object returned in this case has some problems, however, and errors arise if you try to   print the resulting object, although the arrays are set properly I think.

We need a better way to determine the correct `zval` name automatically.  Also some data sets might use something other than `lon` and `lat` for x and y.

mandli:
We do not have `matplotlib` on travis which is why the check is failing.  If you are only using `pylab` to plot than I would just put that in the plotting if-then statement.

PR #310, opened 2018-05-28
Flex max grids

mjberger:
allow maxgr to become larger and create new larger arrays (rnode, node,etc).
also changed gauges_module.
use with amrclaw branch with same name.


mandli:
This looks fine to me pending getting the AMRClaw updates.

Issue #316, opened 2018-06-17
Incompatibility with Master due to timer names


mandli:
The PR https://github.com/clawpack/amrclaw/pull/220 in AMRClaw that fixed some timer bugs has broken GeoClaw due to a conflict in variable names in the new `amr_module` and `tick` in GeoClaw.  Error is below although it is a bit misleading.
```
/Users/mandli/Dropbox/src/clawpack/geoclaw/src/2d/shallow/tick.f:23:30:

/Users/mandli/Dropbox/src/clawpack/geoclaw/src/2d/shallow/tick.f:9:9:

       use amr_module
         2
/Users/mandli/Dropbox/src/clawpack/geoclaw/src/2d/shallow/tick.f:23:30:

       integer tick_clock_start, tick_clock_finish, tick_clock_rate
                              1
Error: Symbol 'tick_clock_start' at (1) conflicts with symbol from module 'amr_module', use-associated at (2)
/Users/mandli/Dropbox/src/clawpack/geoclaw/src/2d/shallow/tick.f:459:24:

       call system_clock(tick_clock_finish,tick_clock_rate)
                        1
Error: 'count' argument of 'system_clock' intrinsic at (1) must be INTEGER
make: *** [/Users/mandli/Dropbox/src/clawpack/geoclaw/src/2d/shallow/tick.o] Error 1
```

rjleveque:
Sorry!  I'll try to fix this later today.

mandli:
I should have caught this too.  Fixed by #317.

PR #317, opened 2018-06-17
remove tick_clock_start declaration from tick.f

rjleveque:
It is now declared in amrclaw/src/2d/amr_module.f90

Addresses #316.


mandli:
This works for me locally, merging.

Issue #309, opened 2018-05-28
Bowl-Radial Multilayer Example Fails


mandli:
@mjberger found a problem with the `filval` routine for multilayer code that involves an invalid value.  @aks2203 can you take a look if you can recreate this and trace down the problem?  It looks like it may be a sqrt of a negative number.

```
[mandli@cu-biotgw-1-10:bowl-radial(master $%)] printenv FFLAGS
-O0 -W -Wall -fbounds-check -fcheck=all -pedantic-errors -Wunderflow -fbacktrace -ffpe-trap=invalid,zero,overflow -g
```

```
[mandli@cu-biotgw-1-10:bowl-radial(master $%)] make output
Makefile:111: warning: overriding commands for target `all'
/Users/mandli/src/clawpack/clawutil/src/Makefile.common:280: warning: ignoring old commands for target `all'
rm -f .output
python /Users/mandli/src/clawpack/clawutil/src/python/clawutil/runclaw.py xgeoclaw                  _output                \
	True None . true False None
==> runclaw: Will take data from  /Users/mandli/Dropbox/src/clawpack/geoclaw/examples/multi-layer/bowl-radial
==> runclaw: Will write output to  /Users/mandli/Dropbox/src/clawpack/geoclaw/examples/multi-layer/bowl-radial/_output
==> runclaw: Removing all old fort/gauge files in  /Users/mandli/Dropbox/src/clawpack/geoclaw/examples/multi-layer/bowl-radial/_output

==> Running with command:
    /Users/mandli/Dropbox/src/clawpack/geoclaw/examples/multi-layer/bowl-radial/xgeoclaw
Reading data file: claw.data
         first 5 lines are comments and will be skipped
Reading data file: amr.data
         first 5 lines are comments and will be skipped

 Running amrclaw ...

Reading data file: geoclaw.data
         first 5 lines are comments and will be skipped
Reading data file: refinement.data
         first 5 lines are comments and will be skipped
Reading data file: dtopo.data
         first 5 lines are comments and will be skipped
Reading data file: topo.data
         first 5 lines are comments and will be skipped

 Reading topography file  /Users/mandli/Dropbox/src/clawpack/geoclaw/examples/multi-layer/bowl-radial/bowl.topotype2
Reading data file: qinit.data
         first 5 lines are comments and will be skipped

 Reading qinit data from file  /Users/mandli/Dropbox/src/clawpack/geoclaw/examples/multi-layer/bowl-radial/hump.xyz

[...]

PR #319, opened 2018-07-25
Filval fineflag error

aks2203:
#309


src/2d/shallow/multilayer/filval.f90
```diff
@@ -236,6 +236,9 @@ subroutine filval(val, mitot, mjtot, dx, dy, level, time,  mic, &
                 ! if all fine cells are dry, momentum has already been set
                 if (finemass >= dry_tolerance(layer)) then
                     do ivar = 3*layer-1,3*layer
+                        ! if (ivar>3) then
+                        !   print *, ivar
+                        ! endif
                         fineflag(ivar)=.false.
```
mandli: Remove these debug lines.

src/2d/shallow/multilayer/filval.f90
```diff
@@ -349,8 +352,8 @@ end subroutine filval
 
 subroutine dumpaux(aux,naux,mitot,mjtot)
    implicit none
-   real(kind=8) :: aux(naux,mitot,mjtot)
    integer :: naux,mitot,mjtot,i,j,iaux
+   real(kind=8) :: aux(naux,mitot,mjtot)
```
mandli: Revert this as it does not actually add anything and we can then check versions easier in the future.

PR #320, opened 2018-07-25
Adding surface deformation model for triangular sub-faults

dsrim:
Added python implementation of surface deformation model due to Comninou and Dundurs, causally referred to as the "triangular version of the Okada model". This allows the computation of surface dislocation due to the slip of a subfault that has triangular shape.

### DIFF
--- a/src/python/geoclaw/dtopotools.py
+++ b/src/python/geoclaw/dtopotools.py
@@ -147,7 +147,8 @@ def plot_dZ_contours(x, y, dZ, axes=None, dZ_interval=0.5, verbose=False,
 
 
 def plot_dZ_colors(x, y, dZ, axes=None, cmax_dZ=None, dZ_interval=None,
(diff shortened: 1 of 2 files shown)


rjleveque:
@dsrim: Some nosetests are failing because the NOAA THREDDS server seems to be down.  I'll test again later.

`test_dtopotools.py` is failing because of the way I redefined `rise_time`, I think.  I will investigate.

rjleveque:
Some checks are failing because of minor changes in dtopo that is being generated.  I'm going to merge this and then work on updating the regression data.

rjleveque:
@dsrim: I also updated some of the documentation to discuss this, in particular http://www.clawpack.org/v5.5.0_future/okada.html.  Does that seem ok?

Eventually we should also add a triangular example to `$CLAW/apps/notebooks/geoclaw/dtopotools_examples.ipynb`

PR #321, opened 2018-07-28
WIP: Triangular Okada updates

rjleveque:
@dsrim: Perhaps I merged #320 prematurely.  I'm having problems getting the the results to look right for a single triangle if it's not oriented just right.

I don't think the strike was always calculated properly before.  If the fault trace runs north-south and dips to the east then strike=0, but if it dips to the west then strike=180, whereas your code was setting it to 0 in both cases.  The proper criterion is that dip should lie between 0 and 90 when the dip is relative to the strike direction.

I think I fixed this in this PR.  However, there is still another problem: If I define the same triangle with the vertices in a different order then I expect the same deformation, I think.  But both your original and my code produce a negated deformation when the vertices are swapped.  Even though all the parameters (strike, dip, rake) are the same.  This is strange.  So it must be in okada() that something I don't understand is being done that makes these come out differently.  Maybe you can figure it out.

Also, if the fault is vertical the original divided by zero and I adjusted so that dip=90 is set in this case.  But then the okada() function fails in this case with error messages.

I posted the notebook I used to experiment with this:
 - http://faculty.washington.edu/rjl/misc/test_triangles.html
 - http://faculty.washington.edu/rjl/misc/test_triangles.ipynb


dsrim:
Thank you for checking and fixing that. As for the negation of the deformation - let me give it a bit more thought.

dsrim:
@rjleveque Reason is the solution (of Comninou and Dundurs) assumes that the corners (vertices) of the triangular subfault should be oriented counter-clockwise. If two vertices are swapped, the orientation reverses, negating the resulting deformation. This can be fixed when we check the sign of ``normal[3]`` (if the normal points downwards, it is oriented clock-wise) by reversing the ordering of given ``self._corners``... I'll first do a PR to your local branch first.

PR #325, opened 2018-08-08
valout.f90 missing line

rjleveque:
eta was printing out as all zeros


mandli:
Not sure how I missed that.  I suppose the tests pass as they are entirely based on gauges?

PR #326, opened 2018-08-09
Adding adjoint flagging method

BrisaDavis:
Adding files and example for adjoint flagging method in GeoClaw. 

This depends on the AMRClaw code found in: https://github.com/clawpack/amrclaw/pull/228


rjleveque:
@BrisaDavis and @mandli:   I tried updating this branch with all the more recent changes to `master`.  The example codes still run fine, but some multilayer tests fail because the changes made to the Makefile and several source routines in `src/2d/shallow` for the adjoint are not properly reflected in the code in `src/2d/shallow/multilayer`.

In particular, `src/2d/shallow/Makefile.geoclaw` includes `$(AMRLIB)/adjoint_module.f90`, which is not included in `src/2d/shallow/multilayer/Makefile.multilayer` but is needed by some routines, so it does not compile.

Also, `multilayer/Makefile.multilayer` still points to `$(AMRLIB)/errf1.f` while `shallow/Makefile.geoclaw` points to `$(GEOLIB)/errf1.f`, a modified version for the adjoint code. Similarly for many other routines, and I'm not sure if this will screw things up in the multilayer code.  

Presumably adjoint flagging isn't going to available in the multilayer code anyway, since the inner products would have to be over longer vectors and it's nontrivial to figure out the right way to make this extension.  But we should make sure the multilayer code still works properly with the other flagging  options, and document the fact that adjoint flagging can't be used.

Note that to try compiling this code, you need to check out `amrclaw` to the `adjoint` branch combined with the recent changes to `master`, as in my recent PR https://github.com/BrisaDavis/amrclaw/pull/7

mandli:
I think it's fine to leave out the multilayer code, maybe we can add an issue for it though.  We should put an error message if someone tries to use it though.

rjleveque:
See https://github.com/BrisaDavis/geoclaw/pull/3 for some discussion of issues we still need to sort out in geoclaw.

Regarding an error message, we should perhaps modify the way `adjoint.data` is created in amrclaw so we can do this kind of check.  Currently creating this is just hardwired into `setrun.py` so we can't do any general checking.  We might want to think more about this before merging https://github.com/clawpack/amrclaw/pull/228.

BrisaDavis:
In regards to the comment in [BrisaDavis#3](https://github.com/BrisaDavis/geoclaw/pull/3) about the new geoclaw errf1.f subroutine having a different argument list than its amrclaw counterpart: since the geoclaw version requires the aux arrays, it seems to me that the correct fix to this would be to update the amrclaw version to add the aux arrays into the argument list even though they wouldn't be used in that subroutine. Provided that errest.f and any other routine that calls errf1.f is updated appropriately, would changing the call sequence for errf1.f in amrclaw have any backwards compatibility issues?

PR #328, opened 2018-08-10
Minor Corrections to `valout` Refactor

mandli:
This mostly brings the GeoClaw refactor into alignment with everything that happened in the AMRClaw refactor.


rjleveque:
Works fine, thanks!

PR #329, opened 2018-08-12
kmltools -- increase number of digits printed

rjleveque:
This increases the number of digits put into kml files showing gauges, and extent of topo, dtopo, regions, both in the number used to position and in the text box descriptions (using a function `f2s` that converts floats to strings with trailing zeros removed but with more digits than the `%g` format - maybe there's a better way to do this?). 

I modified the chile2010 example to make these kml files by default since this is generally very useful to make sure things are placed where expected.  I added some description to the `README.rst`.

In testing this, I also tried 
```
make plots SETPLOT_FILE=setplot_kml.py
```
and discovered that it seems to still make the png files properly (by unzipping Chile2010.kmz) and opening one) but when the kmz file is opened in my current version of Google Earth the tsunami does not show up, nor does the gauge plot.  Maybe @donnaaboise or someone more familiar with GE rendering wants to take a look at this?

### DIFF
--- a/examples/tsunami/chile2010/README.rst
+++ b/examples/tsunami/chile2010/README.rst
@@ -25,15 +25,26 @@ A single gauge captures the sea surface elevation at the location of
 
 **Creating kml files to view on Google Earth**
 
+By default `make data` will create kml files showing the computational
+domain, the extents of the topofiles and dtopofiles used and the
+AMR regions specified, and the gauge location.  These can be viewed
+using Google Earth or other platforms supporting `kml`.  These are
+created by the line in `setrun.py` invoking
+`kmltools.make_input_data_kmls(rundata)`.
+
+Note that clicking on a rectangle or gauge in Google Earth displays
+information about it.  You may need to unselect some of the kml files under
+Places in order to click on ones underneath.
+
 `setplot_kml.py` is a version of the setplot file that produces `kml` files
-suitable for viewing on Google Earth or other platforms supporting `kml`.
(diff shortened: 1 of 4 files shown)


rjleveque:
Yes, that was my problem.  The Levels were turned on so I saw all the patch edges evolving when moving the time slider, but it wasn't showing the tsunami until I clicked on the Sea surface plots.   I seem to recall that in the past everything was turned on by default when opening the kmz file, was that behavior changed recently?

donnaaboise:
There is an attribute that can be set to indicate which figures should show up when GE is first started : 

    plotfigure.kml_show_figure = True

This wasn't set in the chile2010 example.   The regular Visclaw setting `plotfigure.show` is ignored by the KML plotting.

I don't have something equivalent for the gauges, but could easily add  it.

donnaaboise:
I am looking at the default visibility of gauges in the GE plots, and notice that I am using `plotdata.gauges_fignos` to set the visibility. It looks like I was setting it to `0` if `gauges_fignos not None`, and 1 otherwise.  I have fixed this.  

Does it make sense to use the `gauges_fignos` attribute for this?   I don't have a `plotdata.kml_show_gauges` attribute, (which would be a bit more consistent with how I treat the other figures)  but could easily  add it.

rjleveque:
@donnaaboise: I merged in https://github.com/clawpack/visclaw/pull/233. It seems fine for now to use `gauges_fignos`.

I will modify the chile2010 example to plot by default as you suggest in a separate PR.

PR #331, opened 2018-08-21
Change ModuleNotFoundError to ImportError

mandli:

PR #332, opened 2018-08-22
fixing xarray version issue in read_ibtracs and adding NetCDF missing val

bolliger32:
Fixes issue identified in #331.

Also, adds functionality to interpret netcdf _FillValue as missing value and treat it as missing values in other topotypes are treated.


mandli:
@rjleveque Does this fix the issue you were seeing?

src/2d/shallow/topo_module.f90
```diff
@@ -600,6 +602,46 @@ subroutine read_topo_file(mx,my,topo_type,fname,topo)
                     topo(j * mx + 1:(j + 1) * mx) = nc_buffer(:, my - j)
                 end do
                 deallocate(nc_buffer)
+
```
mandli: I think you may have added this on accident.  Not opposed to having this added but wanted to just check in to see what the purpose is.

mandli:
@bolliger32 I think you may have accidentally included some code from your filling topography value stuff.  It would probably be best to put that in another PR although I am in support of having these changes.

bolliger32:
@mandli hmm...only issue is that these changes were made on the same branch (sorry probably not good form). Is there an easy way to submit a PR for only some of the commits on a branch?

src/2d/shallow/topo_module.f90
bolliger32: yes like you alluded to in the PR comment, this is for the netcdf filling missing value functionality.

rjleveque:
This seems to have fixed the problem, thanks @bolliger32.

The best way to do a new PR is to do something like:
```
cd $CLAW/geoclaw
git fetch origin master   # assuming origin points to git://github.com/clawpack/geoclaw
git checkout -b fix_xr_where origin/master  # clean branch based on the real master
git checkout master -- src/python/geoclaw/surge/storm.py  # check out this one modified file
git commit -m "swap xr.where to np.where"
git push bolliger32 fix_xr_where
```
and then do a PR from that branch.  

(For your own sanity it's usually best to make a new branch for any commits you plan to do rather than committing to your master branch, so your master is always up to date with clawpack's master).

I'm going to close this PR so you can try that, or let me know if you run into problems.

The missing value is something we should fix in a more general way regardless of how the topo file is stored, not just for netCDF.  I suggest we leave that until after we get this release out.

bolliger32:
OK thanks! I addressed the xarray issue with a new pr (#334). Will keep the NetCDF stuff as is for now. It does mirror the functionality of the other topo types so that if we did want to give it a more general approach in the future, the netcdf read functionality would be set up to do so. I'm happy to commit that as a separate PR if/when desired.

PR #333, opened 2018-08-23
Update Isaac example to new Storm module and data formats

mandli:
Includes minor tweaks to data fetching (topography and storm data) as well as updating to the newer ways we handle storm data.


rjleveque:
`get_bathy.py` still seems to be missing.

mandli:
It's included in the `setrun.py` instead.

mandli:
- Glad you caught the makefile dependency.  Fixed now.
 - Verbosity has also been changed to `0` for both this and the Ike example.
 - I agree with you on the `get_remote_file` business.  What do we want to use for this?

mandli:
Fixed the issue.  Somehow I had decided that `3,300 == 60^2`.  Who knew that was a thing?

PR #337, opened 2018-09-18
compare q(1,i,j) instead of q(1,i,j)*capa to dry_tolerance in update.f90

xinshengqin:
I would like to raise a discussion on this single line of code in update.f90.

I think here we should compare **q(1,i,j)** instead of **hf = q(1,i,j) * capa** to **dry_tolerance** in update.f90. 

One simple argument is that at all other places, **q(1,i,j)**  instead of **q(1,i,j) * capa** is compared to **dry_tolerance**.

Any thought?

### DIFF
--- a/src/2d/shallow/multilayer/update.f90
+++ b/src/2d/shallow/multilayer/update.f90
@@ -149,7 +149,8 @@ subroutine update (level, nvar, naux)
                                     huf= alloc(iaddf(3*layer-1,iff+ico-1,jff+jco-1,locf,mi))*capa 
                                     hvf= alloc(iaddf(3*layer,iff+ico-1,jff+jco-1,locf,mi))*capa 
 
-                                    if (hf > dry_tolerance(layer)) then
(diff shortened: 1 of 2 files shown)


rjleveque:
Agreed - since q(1,i,j) is the fluid depth this is what should be compared to `dry_tolerance` in general.  The value of `capa` can be huge for lat-long grids (where I believe it's the ratio of cell area in meters to cell area in degrees and hence around 1e10).

mandli:
I also cannot think of any reason that we would have wanted to check the actual mass of water.

Can you make a similar adjustment to the `src/2d/shallow/multilayer/update.f90` value.  Note that `h` in that case needs to keep the division by `rho(layer)`.

mandli:
Could store the value in `hf` without multiplying by `cappa` and then do the multiplication.  Might be a bit superfluous though.  I would say just leave it for now in any case so that the lines look similar.

xinshengqin:
Agree.

PR #338, opened 2018-10-10
improve read_ibtracs method and fix write kwargs bug

bolliger32:
**write**
- kwargs are now passed from `write` wrapper method to `write_geoclaw` and any other `write_*` that may get written in the future. (e.g. passing `max_wind_fill_radius` argument)

**read_ibtracs**
- Using more up-to-date version of v4 BETA release of IBTrACS (slightly different netcdf variable format)
- pull from `usa_wind` and `usa_pres` when `wmo_wind` and `wmo_pres` when latter variables aren't yet available
- set time_offset attribute to final timepoint of ibtracs recording if storm doesn't make landfall (no 'L' event)


mandli:
Not sure what's going wrong here but it seems like it has something to do with the use of a time span?

bolliger32:
whoops. Updated the read_ibtracs code to match the latest beta release of IBTrACS v4, but forgot to update the corresponding netcdf file in the tests directory. Just did that, so hopefully should build OK this time.

bolliger32:
So I already only wrote it for the beta version of ibtracs v4 (which was
supposed to become a public release by September but still hasn't yet).
They must have made a minor formatting change somewhere in the beta process
so it's not backwards compatible with earlier versions of the v4 beta
(which I think is fine) and also not with v3 (which it never was). Do you
think that's ok or should it be compatible with v3 as well?

It also has a lot of todos especially for non-us storms. It doesn't read in
variables from other met agencies yet if they aren't available under to wmo
variables. But I figured something better than nothing at this point :)

mandli:
If the previous code only did v4 then I would not worry about it.  I just did not want to lose functionality.  We may want to put a note in the doc-string that v4 is the version we are supporting.

bolliger32:
cool I did have a note saying that v4 BETA was the only version supported, but I added a quick remark clarifying that it is the v4 BETA version that was current as of 10/8/2018 in case there are more formatting changes before it is released. Supposedly it should be publicly released any day now, so I can update that docstring then.

mandli:
Thanks!

PR #342, opened 2018-11-03
Fix periodic BCs problem with topography

mandli:
Also includes some efficiency changes and clarifications regarding the role of the routine.  This includes:
 - Rewriting the description of the function and adding some notes.
 - Removing extraneous if-statement checks from the interior of loops.  This mostly involves separating out loops that were redundantly checking whether it should set some of the aux fields from within loops.
 - Remove the `forall` statements as they are problematic for debugging and today do not generally lead to speedups.
 - Remove some outdated code.

### DIFF
--- a/src/2d/shallow/setaux.f90
+++ b/src/2d/shallow/setaux.f90
@@ -1,18 +1,33 @@
+! Set auxiliary arrays
+!
+! Routine that sets the values found in the aux array including topography and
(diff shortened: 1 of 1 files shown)


mandli:
Some of this may be translatable to AMRClaw although most of what is done here is specific to GeoClaw.

rjleveque:
Working fine according to @mandli and backwards compatible.

PR #343, opened 2018-11-03
topotools.read_netcdf sometimes failed with coarsen==1

rjleveque:
simplify slicing notation in this case.

@mandli: I think you said at one time that you were having trouble downloading some netcdf topo for  a storm surge problem.  Maybe you can check if this fixes your problem (and/or still works).

### DIFF
--- a/src/python/geoclaw/topotools.py
+++ b/src/python/geoclaw/topotools.py
@@ -1667,6 +1667,9 @@ def read_netcdf(path, zvar=None, extent='all', coarsen=1, return_topo=True,
     if verbose:
         print("Will read netCDF data from \n    %s" % path)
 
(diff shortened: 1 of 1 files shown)


mandli:
I will have to dig up the case that was failing for me but this looks good otherwise.  On a related note while working with @mjberger yesterday I ran into some recurrent issues relating to the names of the `x`, `y` and `z` variables that I think would be easy to add a bit more logic to and make it more pleasant to deal with NetCDF topography files.

mandli:
So this does not exactly fix the problem but I think it helps a lot.  Going to merge.

mandli:
It's more of a constraint to the way we call the THREDDS server so I am not too worried about it.

PR #347, opened 2018-12-10
fix typo in SubFault.dynamic_slip

rjleveque:
very minor so I'll merge it in.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ before_install:
   - pip install xarray
   - git clone --branch=master --depth=100 --quiet git://github.com/clawpack/clawpack
   - cd clawpack
-  - python setup.py symlink-only
+  - python setup.py git-dev
   - rm -rf geoclaw
   - ln -s ../ geoclaw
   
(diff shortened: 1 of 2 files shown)


mandli:
Crazy, how did we not catch that?

mandli:
Looks like the tests failed due to not including one of @ketch's new PRs?

rjleveque:
The `.travis.yml` file still uses `python setup.py symlink-only` which has been deprecated. Is changing this in a PR on geoclaw?  I don't see it.

rjleveque:
@mandli and @ketch: Changing `symlink-only` to `git-dev` on the `python setup.py` line worked, so if this seems like the right way to fix it we can do the same in amrclaw.  It's already doing that in classic.

mandli:
Looks like this works but @ketch would know better than I if this is the right thing.  In PyClaw it looks like a `python setup.py install` is done in PyClaw.

ketch:
I think that's fine. It might be even nicer to just explicitly call git submodule init + update so there is no mystery about what is going on here.

ketch:
In fact I would favor removing the git-dev option for the reason I just stated.

rjleveque:
Is this better?

ketch:
Seems better to me -- no reason to hide two simple and standard commands inside one very non-standard function.

PR #348, opened 2018-12-12
Minor plotting fixes

navravi:
Summary
-------
- Define `surge_data` which was previously being used without being defined
- Make land plots compatible with other `pcolor` plots by not fixing bounds by default; otherwise, the land plot interferes with the scale
- Add name for bathy plotitem

Testing
-------
- Manually tested by making plots for Katrina model


mandli:
These are failing due to a change in our installation options.  It should work if you merge in what's in master right now.

Issue #341, opened 2018-11-03
Change When Gauges are Recorded


mandli:
Currently gauge data is recorded before a time step is taken.  This decision will catch the initial condition but has two drawbacks:
1. the last time point is never captured and
1. if aux fields are not set in `setaux` but somewhere else such as `b4step` these values will be incorrectly recorded every regridding time.

Note that the latter of these is only an issue to how the gauges record aux values and does not impact the actual time stepping.

mandli:
Couple of other thoughts:
 - The primary idea here would be to move the call to `update_gauges` to after `stepgrid` is called. We should also add `update_gauges` to also hit the initial time step.
 - This will break any test that relies on gauges.
 - This should be back-ported to AMRClaw as well.

mandli:
This needs to be checked to see if this works now.

mandli:
I just noticed this was never fully addressed despite the PRs that were merged.  The code that was merged does not seem to be there anymore so it may have been a regression in a commit at some point.  This is a critical bug for gauge recordings.

PR #350, opened 2018-12-12
Fix recording of wind at gauges

navravi:
Summary
-------
This change relates to issue https://github.com/clawpack/geoclaw/issues/341.

Prior to this change, the wind (both x and y components) would be recorded as zero every so often at the gauges, as shown below. The number of time steps between the zero recordings was also not consistent.
```
katrina$ awk 'NR > 2 && NR < 20 {print $11"\t"$12}' _output/gauge00001.txt
0.0000000E+00   0.0000000E+00
-0.7017150E-06  -0.1215794E-05
-0.7017155E-06  -0.1215795E-05
-0.7017781E-06  -0.1215890E-05
-0.7018407E-06  -0.1215984E-05
-0.7019033E-06  -0.1216078E-05
-0.7019659E-06  -0.1216173E-05
0.0000000E+00   0.0000000E+00
-0.7020912E-06  -0.1216361E-05
-0.7021538E-06  -0.1216456E-05
-0.7022165E-06  -0.1216550E-05
-0.7022791E-06  -0.1216644E-05
-0.7023418E-06  -0.1216739E-05
0.0000000E+00   0.0000000E+00
-0.7024671E-06  -0.1216928E-05
-0.7025298E-06  -0.1217022E-05
-0.7025924E-06  -0.1217116E-05
```
The zeros were due to the fact that geoclaw zeroes out the wind field before setting its correct value in `stepgrid`. However, prior to this PR, `update_gauges` was being called before `stepgrid`, so the zeroed out values were being written to the gauges rather than the correct values.

The fix is to move the `update_gauges` call to the end of the `par_advanc` routine, after the call to `stepgrid`.

Notes
-----
Prior to this change, the last time step was not recorded at all, and the first time step was sometimes recorded twice (for storm surge models). After this change, the first time step is sometimes recorded (for storm surge models), and the last time step is always recorded. Issue https://github.com/clawpack/geoclaw/issues/341 mentions that another `update_gauges` call should be added to hit the first time step. However, the fact that the first time step appears to already be recorded for storm surge models needs to be investigated further.

Although this PR changes the behavior in the recording of the first time step, the inconsistent recording of the first time step is an existing bug that should be fixed independently. Calling `update_gauges` before `stepgrid` (the old behavior) was a workaround that ended up introducing a new bug (the wind field being recorded incorrectly).

After this PR, all the time steps are recorded for storm surge models, but the first time step is missing for other models.

The test outputs were overwritten as part of this change.

Testing
-------
After the change, it was manually verified that the wind was being written out without zeros.
```
katrina$ awk 'NR > 2 && NR < 20 {print $11"\t"$12}' _output/gauge00001.txt
-0.7017150E-06  -0.1215794E-05
-0.7017155E-06  -0.1215795E-05
-0.7017781E-06  -0.1215890E-05
-0.7018407E-06  -0.1215984E-05
-0.7019033E-06  -0.1216078E-05
-0.7019659E-06  -0.1216173E-05
-0.7020286E-06  -0.1216267E-05
-0.7020912E-06  -0.1216361E-05
-0.7021538E-06  -0.1216456E-05
-0.7022165E-06  -0.1216550E-05
-0.7022791E-06  -0.1216644E-05
-0.7023418E-06  -0.1216739E-05
-0.7024044E-06  -0.1216833E-05
-0.7024671E-06  -0.1216928E-05
-0.7025298E-06  -0.1217022E-05
-0.7025924E-06  -0.1217116E-05
-0.7026551E-06  -0.1217211E-05
```

Output of `nosetests`:
```
geoclaw (gauges)$ FFLAGS=-llapack nosetests
...........S...S/home/nav/claw/clawpack/clawpack/geoclaw/surge/storm.py:451: UserWarning: *** Cannot yet automatically determine the
    maximum wind radius.  Will write out GeoClaw
    formats but note that these will not work
    when running GeoClaw currently without a custom
    `max_wind_radius_fill` function passed as argument
    to the `write` function.
  warnings.warn(missing_data_warning_str)
/home/nav/claw/clawpack/clawpack/geoclaw/surge/storm.py:643: UserWarning: *** Cannot yet automatically determine the
    maximum wind radius.  Will write out GeoClaw
    formats but note that these will not work
    when running GeoClaw currently without a custom
    `max_wind_radius_fill` function passed as argument
    to the `write` function.
  warnings.warn(missing_data_warning_str)
...S...S.....
Ran 29 tests in 52.915s

OK (SKIP=4)
```


mandli:
See #348

mandli:
As with clawpack/amrclaw#235 this changes the regression data due to the modification of what is being recorded by gauges.  This looks good to me but another should check before merging.

mandli:
This is being superceded by #357.

PR #351, opened 2018-12-21
switch dx to dy in qinit for bowl-slosh example

wmayfield:


mandli:
Thanks for catching that!  @wmayfield Can you also make the same fix in the bowl-slosh test?

wmayfield:
Sure :-)

PR #353, opened 2019-01-23
Changes of storm.py for JMA files

RunxinNi:
1. Revised_read_jma is codes to add radius of maximum wind data for JMA files.
2. calculate_radius_max_wind is codes of a function to calculate radius of maximum wind for JMA files.


Revised_read_jma
```diff
@@ -0,0 +1,26 @@
...
+                                                        'km', 'm')
+            # Warning for formats that have yet to have a default way to determine crticial
+            # storm radius from the input data
+            missing_data_warning_str = """*** Cannot yet automatically determine the storm radius.
+                Will write out GeoClaw formats but note that these will
+                not work when running GeoClaw currently."""
+            warnings.warn(missing_data_warning_str)
+            
+            # Current storm radius is not given
+            self.storm_radius[i] = 0
```
mandli: This should be set to `-1` instead of `0`.

Revised_read_jma
```diff
@@ -0,0 +1,26 @@
+            # Mark if this is a shortened line - does not contain 30 and 50 kts
+            # wind radius - set those to -1 to mark them as missing
+            if len(data) < 8:
+                self.max_wind_radius[i] = -1
+            else:
+                data[7] = int(data[7]) - int(int(data[7])/10000)*10000
```
mandli: Try to adhere to the PEP8 code-styling guidelines.  Here putting spaces around operators is what is important but please take a look at the guidelines for other ideas.

Revised_read_jma
```diff
@@ -0,0 +1,26 @@
...
+                data[9] = int(data[9]) - int(int(data[9])/10000)*10000
+                #data[8] = (data[7] + int(data[8]))/2
+                #data[10] = (data[9] + int(data[10]))/2
+                self.radius_30kts[i] = units.convert(float(data[9]), 'nmi', 'km')
```
mandli: I am interested in the idea of storing these values rather than keeping them locally.  Have you properly initialized the arrays in the `__init__` method though?

Revised_read_jma
```diff
@@ -0,0 +1,26 @@
+            # Mark if this is a shortened line - does not contain 30 and 50 kts
+            # wind radius - set those to -1 to mark them as missing
+            if len(data) < 8:
+                self.max_wind_radius[i] = -1
+            else:
+                data[7] = int(data[7]) - int(int(data[7])/10000)*10000
+                data[9] = int(data[9]) - int(int(data[9])/10000)*10000
+                #data[8] = (data[7] + int(data[8]))/2
```
mandli: Remove commented out lines.

calculate_radius_max_wind
```diff
@@ -0,0 +1,60 @@
+def calculate_radius_max_wind(rho_air, latitude, max_wind_speed, central_pressure,
+                              ambient_pressure, radius_30kts, radius_50kts):
+    """
```
mandli: You should still provide a brief description of what this is doing and perhaps some references to papers it is based on.  Also some of what the input is may be in a storm object which would shorten the number of input arguments.

calculate_radius_max_wind
```diff
@@ -0,0 +1,60 @@
...
+    - *ambient_pressure* (float) Ambient pressure.
+    - *latitude* (flloat) Latitude of center of storm.
+    - *max_wind_speed* (float) Speed of maximum wind.
+    - *central_pressure* (float) Central pressure of storm.
+    - *radius_30kts* (float) Maximum radius of 30 kts wind.
+    - *radius_50kts* (float) Maximum radius of 50 kts wind.
+
+
+    OUTPUT:
+    - *radius* (float) Radius of maximum wind.
```
mandli: So this only computes it for one value?  Maybe we should have the ability to compute for every time point in a `storm` object?

Revised_read_jma
RunxinNi: And about storing the value, I convert data[7] and data[9] into correct format and store them in self.radius_30kts and self.radius_50kts. And I initialized the arrays in the __init__ method though.

Revised_read_jma
RunxinNi: But if we set it to 0, there will be nothing appear in storm data document. Could we consider to set it to -1?

Revised_read_jma
RunxinNi: Ok, I will revise it.

Revised_read_jma
RunxinNi: Ok, I will remove it.

calculate_radius_max_wind
RunxinNi: I seem to get what your mean.

RunxinNi: Ok, I got it!

mandli:
There's something weird about this PR, it looks like the history thinks that you are adding a whole new `storm.py`.  Can you try and maybe rebase this on the current version or try copying your changes back onto what is now the master branch?  Basically you need to resolve the conflicts with the master branch so that the above thing is green.

mandli:
A couple of comments:
 - Given that you are adding the Mangkhut example to surge-examples we should probably not include it here.
 - The new JMA test file is not being read in correctly, hence the failed checks.  Can you check to see why?
 - `wind_intensity` probably should be renamed `wind_speed` to be coherent with the `max_wind_speed` value and should be in m/s rather than knots.

RunxinNi:
Hi Kyle, thank you for your comments! I completed checking new JMA test file. I make the change in the `read_jma` in storm.py. I change `num_lines = int(data[12:14])` into `num_lines = int(data[13:15])` and `AssertionError` disappeared. But it seems weird that it shows error in `check_geoclaw` `numpy.testing.assert_almost_equal(data[0], data[1])`.

mandli:
If you changed the data that the JMA test was using then the GeoClaw test will also break.

RunxinNi:
Hi Kyle, thank you for your instructions! I updated jma.txt to original file and updated storm.py. Now all checks have passed.

src/python/geoclaw/surge/storm.py
```diff
@@ -147,6 +147,10 @@ class Storm(object):
      - *time_offset* (datetime.datetime) A date time that as an offset for the
        simulation time.  This will default to the beginning of the first of the
        year that the first time point is found in.
+     - *wind_speed* (ndarray(:)) Wind speed defined in every record, such as 34kt,
+       50kt, 64kt, etc. Default units are meters/second.
+     - *wind_speed_radius* (ndarray(:)) Radii of every record wind speed. 
+       Default units are meters.
```
mandli: I would suggest changing this to be a 2D array with one column holding the radii and the other the speeds.  It can then just be called `wind_speeds`.

RunxinNi: Hi Kyle, I made the requested change.

PR #354, opened 2019-02-24
Geoplot deprecated

rjleveque:
Add deprecated warning to `geoplot.py`
    
All functionality is duplicated in `clawpack.visclaw.geoplot` after clawpack/visclaw#238.


mandli:
I have not been able to replicate this error, it may be due to some incompatibility with version on travis maybe?  @bolliger32 do you have an idea?  The problem is with the ibtracs conversion test.

bolliger32:
Sorry @mandli what is the error you're referring to?

src/python/geoclaw/geoplot.py
```diff
@@ -1,14 +1,23 @@
...
 from clawpack.visclaw import colormaps
 from matplotlib.colors import Normalize 
 from clawpack.geoclaw import topotools
 from numpy import ma
 from six.moves import zip
+import warnings
+
+w = """
+*** WARNING: clawpack.geoclaw.geoplot is deprecated
+             Use clawpack.geoclaw.geoplot instead"""
```
bolliger32:
```suggestion
             Use clawpack.visclaw.geoplot instead"""
```

mandli:
The Travis tests are failing.

bolliger32:
> The Travis tests are failing.

Ah right. Yeah that's weird as nothing really changed here. The current master branch passes travis tests, no?

bolliger32:
Yeah @mandli I'm unable to replicate this error as well. Not sure what's going on. I'm guessing some netcdf incompatibilities.

bolliger32:
looks like `xarray` changed minor versions from `0.11.0` to `0.11.3` since the last time the master branch built successfully. Seems like there might be an `xarray` bug. I confirmed that it builds successfully if you pin `xarray==0.11.0` (but not >= 0.11.1)

mandli:
So installing xarray `0.11.0` seems to fix the problem on travis although I also cannot replicate the problem if I upgrade to `0.11.3` on my own machine.  In any case we should add to the `.travis.yml` file `pip install xarray==0.11.0 for the time being.  I will open up another PR to do that.

bolliger32:
> So installing xarray `0.11.0` seems to fix the problem on travis although I also cannot replicate the problem if I upgrade to `0.11.3` on my own machine. In any case we should add to the `.travis.yml` file `pip install xarray==0.11.0 for the time being. I will open up another PR to do that.

Interesting. Yeah I was just about to try to see what happens if I try 0.11.13 on my machine...Might be some interaction with xarray and netcdf, if netcdf is installed differently on your machine and the travis machine? I wonder if we would see the error if we create a docker container that's built exactly as `.travis.yml` file specifies and run the tests there. I'll admit I'm a travis novice so I'm a little out of my element here :)

Makes sense to pin `xarray` as the temp fix at least

bolliger32:
FYI @mandli looks like this bug was fixed in `xarray>=0.12` in case we want to change the travis recipe. But could also just leave it at `xarray==0.11.0`

mandli:
If travis grabs 0.12 then it would be best to use the newest I would say.

PR #362, opened 2019-04-02
Update .travis.yml to reflect fixed version of xarray

mandli:


bolliger32:
@mandli We should close this as it is fixed by #363

mandli:
As suggested closed in lieu of #363.

PR #363, opened 2019-04-02
Update package dependencies for travis and tests

bolliger32:
- pull latest xarray version on travis  now that latest is `>=0.12` and the bug we had with `0.11.3` no longer requires us to pin `xarray==0.11.0` (see #362 and #354)
- install `scipy` on travis so that no tests requiring `scipy` are skipped
- remove unneeded import of `pylab` package that causes `test_topotools.test_against_old()` to be skipped if `pylab` not installed. `pylab` is no longer needed for that test anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,13 +8,10 @@ env:
 before_install:
   - sudo apt-get update
   - sudo apt-get install gfortran liblapack-pic liblapack-dev
-  # Print NumPy version that is already installed by Travis CI:
-  - python -c "import numpy; print(numpy.__version__)"
-  # - pip install matplotlib
-  # - python -c "import matplotlib; print(matplotlib.__version__)"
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -8,12 +8,9 @@ env:
 before_install:
   - sudo apt-get update
   - sudo apt-get install gfortran liblapack-pic liblapack-dev
-  # Print NumPy version that is already installed by Travis CI:
-  - python -c "import numpy; print(numpy.__version__)"
```
mandli: I think keeping this for debugging purposes is still a good idea (unless it does not work).

bolliger32: ok cool. Nope it works. I moved it to the end of `before_install` just in case the other packages cause numpy to get updated.

PR #366, opened 2019-04-11
prevent underflow in set_holland_1980_fields

delgadom:
surge/model_storm_module.f90:set_holland_1980_fields

current implementation can throw an underflow error if `-Wunderflow` compiler flag is set, as `exp(-(mwr/r)**B)` can be an extremely small value.

```fortran
! Set pressure field
aux(pressure_index,i,j) = Pc + dp * exp(-(mwr/r)**B)

! Speed of wind at this point
wind = sqrt((mwr/r)**B &
        * exp(1.d0 - (mwr/r)**B) * mws**2.d0 &
        + (r * f)**2.d0 / 4.d0) - r * f / 2.d0
```
Limiting the ratio `(mwr/r)` to be less than 10 prevents this error while not significantly affecting wind or pressure estimates.

```fortran
! use a threshold of 10 to prevent underflow
IF ((mwr/r) > 10) THEN
    ! Set pressure field
    aux(pressure_index,i,j) = Pc

    ! Speed of wind at this point
    wind = sqrt((r * f)**2.d0 / 4.d0) - r * f / 2.d0
ELSE
    ! Set pressure field
    aux(pressure_index,i,j) = Pc + dp * exp(-(mwr/r)**B)

    ! Speed of wind at this point
    wind = sqrt((mwr/r)**B &
            * exp(1.d0 - (mwr/r)**B) * mws**2.d0 &
            + (r * f)**2.d0 / 4.d0) - r * f / 2.d0
END IF
```

### DIFF
--- a/src/2d/shallow/surge/model_storm_module.f90
+++ b/src/2d/shallow/surge/model_storm_module.f90
@@ -1,12 +1,12 @@
 ! ==============================================================================
-! model_storm_module 
+! model_storm_module
 !
 ! Module contains routines for constructing a wind and pressure field based on
-! the a parameterized model of the wind and pressure fields.  
-! 
+! the a parameterized model of the wind and pressure fields.
+!
 ! ==============================================================================
 !                   Copyright (C) Clawpack Developers 2017
-!  Distributed under the terms of the Berkeley Software Distribution (BSD) 
+!  Distributed under the terms of the Berkeley Software Distribution (BSD)
 !  license
 !                     http://www.opensource.org/licenses/
 ! ==============================================================================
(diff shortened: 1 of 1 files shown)


mjberger:
Would't you prefer  set to zero, and no trigger of underflow, rather than setting a limit?
I set the traps for everything  EXCEPTN an underflow.  Why do that?


 — Marsha

mandli:
One of the problems we ran into was a bus error that APPEARS to be caused by the underflow of the evaluation of the `exp` function.  One thing I had perhaps thought about doing instead was to do test the argument to the `exp` function and then set it to zero if the argument was small enough.

mjberger:
Did you test on other compilers?

 — Marsha

mandli:
On the list...  The `gfortran` compiler did catch the underflow exception when enabled though.

mandli:
@delgadom are we still having issues with this?

delgadom:
@mandli sorry to drop this - thanks for the ping! We haven't run with the `-Wunderflow` flag since April - will check and get back to you.

delgadom:
If I'm remembering correctly, the intel compiler did not raise an error when you ran this, @mandli. We ran into the issue because we're using gfortran.

mandli:
Yeah, I figured.  We should probably try it again and see if it's a particular version of gfortran.

mandli:
Going to close this as it seems to be compiler (version) dependent for the moment.  If this ends up becoming an issue we can reopen.

PR #368, opened 2019-04-12
Further expand the timing csv header line

mandli:


rjleveque:
As in https://github.com/clawpack/amrclaw/pull/241, please also increase string length for headers in `amr2.f90`.

mandli:
Done.

PR #369, opened 2019-04-13
Update read ibtracs

bolliger32:
- read_ibtracs now handles storms monitored by any agency, and uses those agencies to fill in wind, pressure, and radii data for intervals that are not filled by wmo variables (i.e. uses `usa_wind` to fill in 3hr intervals between the 6hr reported `wmo_wind`)
- updated the test output for `read_ibtracs` accordingly
- I'm not sure why it's saying I changed `.travis.yml`. As far as I can tell, it's the same as in the current commit in master.


mandli:
I think you have changes that revert from the master.  If you rebase on the master branch those should go away.

bolliger32:
not sure what happened but rebasing isn't working. I'm closing this in favor of #370 which I don't think has the revert problem

PR #371, opened 2019-04-15
add QA/QC module and instability checker function

bolliger32:
See docstring for info on how the function works. This will definitely be useful for the Climate Impact Lab work, so if you don't think this is a good general function, we're happy to just keep it on our fork. Otherwise, feel free to merge! I'm imagining other potential ways to do QA/QC that could be added to this module in the future. Also happy to rearrange the structure/move to another module if you think there's a better way to structure it.

### DIFF
--- a/src/python/geoclaw/surge/quality.py
+++ b/src/python/geoclaw/surge/quality.py
@@ -9,13 +9,13 @@ import numpy as np
 class InstabilityError(Exception):
     pass
 
-def check_model_stability(model_output_dir,
-                          regions_to_check = [[-81,22,-40,55], # atlantic
-                                              [-100,15,-82,32], # gulf
-                                              [-88.5,13.25,-70,19.75]], # carribean
-                          frames_to_check = 4,
-                          mean_tol_abs = .01,
-                          min_depth = 300):
+def quality_check(model_output_dir,
+                  regions_to_check = [[-81,22,-40,55], # atlantic
+                                      [-100,15,-82,32], # gulf
+                                      [-88.5,13.25,-70,19.75]], # carribean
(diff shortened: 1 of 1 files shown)


src/python/geoclaw/surge/quality.py
```diff
@@ -0,0 +1,84 @@
...
+            # adjust for sl_init
+            all_vals = all_vals - sl_init
+
+            if all_vals.shape[0]>0:
+                if abs(np.nanmean(all_vals)) > mean_tol_abs:
+                    raise InstabilityError("Model possibly unstable due to large magnitude deep "
+                                           "ocean surge at end of run ({:.1f} cm in region {})".format(
+                                           np.nanmean(all_vals)*100, r))
+
+    return None
```
mandli: Probably do not need this `return` here unless I missed somewhere that it is also called.  I also might suggest naming the function `quality_check` or something other than stability.

mandli:
I think this could definitely be useful and maybe even beyond the storm surge case but I think it would take more time to figure out how to generalize it.

src/python/geoclaw/surge/quality.py
bolliger32: cool. Changed name to `quality_check` and removed the extraneous `return`. And agreed on the generality - I don't think I know enough about the structure of non-geoclaw use cases of clawpack to make this more general, so maybe we keep it within the geoclaw submodule for now and think about how to make something like this more general.

mandli: The more general applications would still be GeoClaw based but they still would be slightly different depending on if it was for tsunamis, debris-flows, dam-breaks, etc.

PR #373, opened 2019-04-21
Adjoint with new q-wave Riemann solvers

rjleveque:
WIP - This is an updated version of #326 with new Riemann solvers that might work better, but still being tested.  For now the new solvers are in `src/2d/shallow/rp*2_geoclaw_adjoint_qwave.f90` rather than in the `riemann` repository of easier testing and updating, but eventually they should be moved.

The adjoint test problem and regression data have also been updated.  I'm curious if travis tests now pass.  (I'm having problems getting an unrelated test to pass on my laptop due to netcdf4/xarray problems.)

More testing needs to be done on the adjoint solvers, will work with @BrisaDavis on that.


mandli:
Removing the `FFLAGS` from `tests/chile2010_adjoint/adjoint/Makefile` as well seems to correct the surge and multilayer issues.  There are still additional problems though with the pathing.

mandli:
Now looking further at what the `chile2010_adjoint` test case is doing I think it may be more prudent to store the forward test data rather than generate it.  Thoughts?

rjleveque:
Travis still fails to compile dtopo1 and multilayer examples for some reason.

@mandli: did you mean we should store the adjoint solution rather than regenerate it?  At the moment it generates 5 frames of data, about 12M in size.  

Maybe best would be to have a separate test that runs the adjoint and saves the data in a place the forward test can find it, and make sure they are run in the proper order first.   

But I'm still unsure why tests pass in amrclaw when the setup there for the adjoint test is the same as here?

mandli:
I was thinking of storing the forward runs and only regenerate the adjoint runs.

rjleveque:
But the adjoint is run first and then snapshots of the adjoint are used to guide refinement in the forward run.

rjleveque:
@mandli: Note travis does pass if I omit these tests, so they were still keeping other codes from compiling properly.  Will investigate further, but we should try to merge something soon since the master branch of geoclaw is currently broken.

mandli:
I am working to provide a version of the `regression_tests.py` file that should run the forward and backward versions of the code and not require explicit `Makefile` calls.  The one thing I did notice was that the Riemann solvers are in the `src/2d/shallow` directory rather than in the Riemann repository.  Why not put this in with the other GeoClaw Riemann solvers?

mandli:
So I have constructed (both not entirely tested) the snipped in this [gist](https://gist.github.com/mandli/3342ae02098e3a7c88351d60cdc1bd00).  Note that the main idea is to have the forward and adjoint runs be done by the same test class, the forward is done using the usual structures and the backward is done with some additional lines of code.  These still need to point to some of the files correctly and the tests themselves will need to be redirected but I think this will avoid some of the issues that are currently breaking things.

rjleveque:
Thanks @mandli, I'll experiment with these a bit.

I put the Riemann solvers in here for now because we're still potentially debugging those and it's easier  to not have to update multiple repositories and clawpack/clawpack to test things on travis.  But eventually they should be moved.

rjleveque:
Currently several tests fail on my laptop because the NCEI netCDF DAP server is down.  This is something we should handle gracefully in our tests, since it seems to happen with some regularity.

rjleveque:
After discussing with @mandli, we decided to merge this for now so other things in geoclaw will work again on master.  I created an issue #377 listing some things still to do on the adjoint code.

Also tests will currently fail due to NCEI server still being down, see #376.

PR #375, opened 2019-04-23
handle missing ATCF pressure

bolliger32:
Some ATCF data files for old storms don't have central pressure. Right now, this will throw an error trying to read and convert an empty string. This PR fixes that and assigns a -1 missing value. It also updates `write_geoclaw` such that it won't print a line with a -1 value for `max_wind_speed` or `central_pressure`. We could treat these similarly to the radius parameters and allow for a "fill function" but for now I think it's fine to just drop these rows?

### DIFF
--- a/src/python/geoclaw/surge/storm.py
+++ b/src/python/geoclaw/surge/storm.py
@@ -352,9 +352,15 @@ class Storm(object):
             else:
                 self.eye_location[i, 0] = -float(data[7][0:-1]) / 10.0
 
-            # Intensity information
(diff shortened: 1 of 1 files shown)


mandli:
Waiting for issues to be resolved in the adjoint merge.

bolliger32:
accidentally deleted this branch but restored it.

mandli:
Is this ready to be merged then?

bolliger32:
yup! long as it looks good to you. It is just for catching the case where you have old storms from ATCF that are missing important things like central pressure or max wind speed

mandli:
I just wanted to make sure that it was not WIP.  All looks good to me otherwise.

PR #379, opened 2019-04-29
move friction setting to after bathy

### LINKED ISSUE #378: variable friction not applied to amr level 1
variable friction is only applied to amr levels > 1

bolliger32:
(hopefully) fixes #378 

I just moved the `set_friction_field` call to after the bathy is set. Confirmed that this fixes the issue where variable friction wasn't applied in amr level 1, but I don't know enough about the fortran source to know if this will create other problems...so it's worth some thought from a more sophisticated geoclaw mind :)

also I don't know why it's saying I changed the whole file. If you know, let me know and I can fix it


bolliger32:
I'm actually surprised that this didn't mess up any of the tests...

bolliger32:
ah it looks like this is because only the Ike regression test incorporates variable friction and none of the gauge files that it compares results to actually experience any surge (eta=sl_init) for all times for each of the 4 gauges checked. Might be worth picking new gauge locations so they get some surge.

mandli:
Any idea why the entire file looks like it's different?

bolliger32:
No haha no idea. I can try with a different text editor. gimme a sec

bolliger32:
oof my bad - will close and open another PR instead

PR #386, opened 2019-05-19
WIP: Test travis adjoint

rjleveque:
Modify tests/chile2010_adjoint/regression_tests.py to copy adjoint subdirectory and run that code as part of setUp.

Contains some debug print statements too.

### DIFF
--- a/tests/chile2010_adjoint/regression_tests.py
+++ b/tests/chile2010_adjoint/regression_tests.py
@@ -64,7 +64,7 @@ class Chile2010AdjointTest(test.GeoClawRegressionTest):
         print('+++ Running forward in directory ',os.getcwd())
         print('+++   contents: ', os.listdir())
         os.system('python maketopo.py')
-        print('+++   scratch directory: ', scratch_dir
+        print('+++   scratch directory: ', scratch_dir)
         print('+++   contents of scratch: ', os.listdir(scratch_dir))
         os.chdir(start_dir)
 
--- a/tests/chile2010_adjoint/regression_tests.py
+++ b/tests/chile2010_adjoint/regression_tests.py
@@ -18,6 +18,15 @@ import clawpack.geoclaw.test as test
 import clawpack.geoclaw.topotools as topotools
 from clawpack.clawutil.test import wip
 
+
+try:
+    CLAW = os.environ['CLAW']
+except:
+    raise Exception("*** Must first set CLAW enviornment variable")
+
+# Scratch directory for storing topo and dtopo files:
+scratch_dir = os.path.join(CLAW, 'geoclaw', 'scratch')
+
 thisfile = os.path.realpath(__file__)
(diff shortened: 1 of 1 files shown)


mandli:
I am getting the following error
```
Program received signal SIGFPE: Floating-point exception - erroneous arithmetic operation.
```
when running `BowlSlosh` test.  Can you reproduce this?  I am using the following flags for `gfortran`:
```
FFLAGS = -O0 -W -Wall -fbounds-check -fcheck=all -pedantic-errors -Wunderflow -fbacktrace -ffpe-trap=invalid,zero,overflow -g
```

rjleveque:
No I can't reproduce, that example runs fine for me with those flags.

mandli:
Hmm, let me take another crack at it then.

mandli:
Ok, not getting that error anymore (not sure what was going on).  I am now getting the same error that Travis is getting.  Attached is the gauge file, it looks like it only differs a bit.
[gauge00001.txt](https://github.com/clawpack/geoclaw/files/3200230/gauge00001.txt)

rjleveque:
@mandli, can you compare what you get from running the adjoint code, in 
```
tests/chile2010_adjoint/adjoint/_output/gauge00001.txt
```
 with what's in 
```
tests/chile2010_adjoint/adjoint/regression_data/gauge00001.txt
```
which is also at:
https://github.com/rjleveque/geoclaw/blob/test_travis_adjoint/tests/chile2010_adjoint/adjoint/regression_data/gauge00001.txt

I'd like to know whether the adjoint solution seems to be the same and the difference only appears in the forward solver, or if the adjoint solution is different.

mandli:
As far as I can tell the adjoint gauges match.  You may want to add some code though to also copy the adjoint results into the directory copied locally if failure is reached.  I ended up running the adjoint code directly to compare so I am not entirely sure if I am comparing what is found in the test.

mandli:
Nevermind, you are copying the test data.  I reran it just to be sure but I diffing the gauges from the adjoint problem reveals nothing.

mandli:
Do you want to remove the `@wip` on the test now that it is passing?

rjleveque:
@mandli: I diffed things from the tar file you sent me and the adjoint solution was identical, but for some reason the forward solution had variations in the 6th decimal place or so although the all the grids created and output times are identical.

I noticed from `_output/claw_git_status.txt` that you had an earlier branch `adjoint_qwave_updates` checked out rather than this update, but when I checked out that branch and ran the test I still get the same results that match the regression data.

So the tests now pass because I changed `rtol` and `atol` to `1e-5`.  I think at this point we should merge it in and go ahead with the release.

But after merging this I want to move the new Riemann solver to the riemann repository and then update clawpack/clawpack and hope it still works.

mandli:
Sound good.

PR #388, opened 2019-05-26
Fix adjoint regression data

rjleveque:
The earlier problems in #386 and #384 was due to the fact that when I ran this test on my laptop it was using a stale version of `scratch/dtopo_usgs100227.tt3` and I think we tweaked the Okada model in the meantime so fresh tests on travis or elsewhere generated slightly different dtopo and hence different gauge output.


rjleveque:
Hurray!  I'll merge this to update https://github.com/clawpack/clawpack/pull/154

PR #394, opened 2019-06-09
Add back square basin example

mandli:
This adds an example for idealized tests including synthetic storm creation.


mandli:
Renamed the example to synthetic given that the basin in question was really not square.

PR #396, opened 2019-06-19
Make a Faster `filpatch`

mandli:
Related to clawpack/amrclaw#243.


mandli:
I think this is now failing for some reason due to AMRClaw but I have not traced it.  Working on it.

PR #398, opened 2019-07-08
Fasterfilpatch

mandli:
Integrate the new `filpatch` changes into GeoClaw.


rjleveque:
This works for me on a few test tsunami problems.  The results disagreed at the 1e-5 level in the case I looked at carefully, but the resulting plots were bitwise identical.

It seems to speed up the BC/ghost cell part of the timing by a few seconds but when I ran problems repeatedly on my laptop, the faster_filpatch version was often slower than the original because variations in timings of other parts of the code were much larger than the savings in filpatch.  Maybe there are other problems where the advantage is more obvious.

At any rate, seems ok to merge.

mandli:
I am surprised that you are seeing differences.  Maybe this is because we are not recomputing the bathymetry as much?  Not sure how that could actually happen though.

In any case we have found a 2x speed up in another storm surge test case.  I assume that with additional aux array entries this may be more advantageous as well.

PR #399, opened 2019-07-23
fix fgmax_read.f90 check for missing file

rjleveque:
Need to inquire if the file exists before opening it.
There was an extra un-needed open statement in the wrong place.


mandli:
Not sure why this is showing the checks have not completed but this is now passing.

Issue #296, opened 2018-05-04
Support reading subset or subsampled topo from NetCDF files


rjleveque:
Improve `topo_module.f90` to allow specifying that only a subset or a strided subset of data from a NetCDF DEM should be saved in a `topo` array.  This would allow, for example, pointing to the same very large 1/3" NetCDF file regardless of what small region is needed for a particular run, rather than creating many different small DEMs for each community. Coarsening the data on input would allow smaller topo arrays and quicker computation of cell-averaged topo in regions where the full resolution is not required.

I think we can use the `NF_GET_VARA` command in Fortran for this.

This would also require modifying `rundata.topo_data` to allow specifying not only the path to the file but also the extent and stride desired.  I will raise a separate issue with suggestions for making this and other `rundata` inputs more flexible.

rjleveque:
#400 is one step in this direction.

It would also be very useful to be able to read in, for example a 1/3 arcsecond DEM covering all of Puget Sound and specify that over one subset we need this full resolution, while over the rest of the Sound we only need 2 second resolution, so please subsample by 6 in each direction.  We currently do this by making separate subsampled DEM files.  For our work with WA State we have to look at several different study regions so the 1/3" data is needed in a different area for each set of runs.

This could be done by reading in the same DEM twice, once specifying the smaller extent with coarsen=1, and once specifying the larger extent with coarsen=6, and internally store the two different topo arrays as if they had been read from different topofiles.  It looks like #400 supports the first of these.

PR #400, opened 2019-07-28
Only read part of netCDF inside domain

bolliger32:
This is useful particularly for the use case where you have a global DEM that you want to repeatedly use as the lowest-resolution "base" topography for multiple storm runs, but you don't want to read in the entire world of topography if each domain is much smaller.

### DIFF
--- a/src/2d/shallow/topo_module.f90
+++ b/src/2d/shallow/topo_module.f90
@@ -974,7 +974,7 @@ contains
         end select
 
         close(iunit)
-        
+
         write(GEO_PARM_UNIT,*) '  mx = ',mx,'  x = (',xll,',',xhi,')'
         write(GEO_PARM_UNIT,*) '  my = ',my,'  y = (',yll,',',yhi,')'
         write(GEO_PARM_UNIT,*) '  dx, dy (meters/degrees) = ', dx,dy
--- a/src/2d/shallow/topo_module.f90
+++ b/src/2d/shallow/topo_module.f90
@@ -18,7 +18,7 @@
 ! ============================================================================
 module topo_module
 
-    use amr_module, only: tstart_thisrun
+    use amr_module, only: xlower,xupper,ylower,yupper
     implicit none
 
     logical, private :: module_setup = .false.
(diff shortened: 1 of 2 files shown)


mandli:
Are you up to possibly modifying the storm surge test case to utilize this feature with a downloaded NetCDF file?

bolliger32:
@mandli sure. You mean the test under `storm_surge`, which currently uses an ascii topo?

mandli:
Yes, if you can do it easily.

bolliger32:
Doable, but the easier thing to do might be to adjust the netcdf_topo test? We could just make the `bowl.nc` much bigger. Do you want this to be done specifically with a downloaded netcdf or would the bowl option work? Also, how were you thinking we'd test that it only read in the right amount of data? I think that might be hard. Or would you just want to maybe save two .nc files (one the correct size for the domain and one bigger) and then test that the outputs are the same? That would be more doable

mandli:
Whatever is easiest to get at least a “not fail” test in there.

bolliger32:
ok just made the bowl.nc file have a bigger range in the netcdf_topo test. This is confirming that the model results are the same, but doesn't necessarily confirm that the new code is working to only load in a subset of the data. To confirm that I think we'd need to monitor memory or somehow add a flag within the fortran code to save the extent of the netcdf that was read, no? That seems probably overkill to me but if you think it's important I'm happy to take that on. I think first order test, however, is to make sure we didn't change the results, which I think is accomplished with just making the bowl.nc file have a larger extent and verifying that the tests still pass.

mandli:
I think if we can confirm it working once (memory wise maybe) then as long as it continues to not break we are good.  Worst case scenario is that it cuts too far.

bolliger32:
I have confirmed it works memory-wise on our runs. And I think the fact that the tests still pass (including gauge results) suggests that nothing is cutting too far? Happy to implement other tests if you have some in mind though

mandli:
Nope, I think this looks good.

Also did you know that shift-cmd-return will comment and close a PR?

bolliger32:
> Also did you know that shift-cmd-return will comment and close a PR?

Haha, I do now!

mandli:
...and cmd-return just comments.  The wonders of keyboard short-cuts.

rjleveque:
Haven't tried it yet, but looks very useful. Thanks!

There are a number of related things I would love to be able to do. See #296.

PR #401, opened 2019-07-30
handle missing storms in read_ibtracs

bolliger32:
In our use cases, we found a need to not throw an error if you feed it a storm that actually has no observed max wind or central pressure values, but to instead return a `Storm` object with no time points. I figure this is a consistent behavior so maybe worth merging into master. But if you think it's better to throw a ValueError in these situations, I'm happy to keep this change on `ClimateImpactlab/geoclaw` and not merge.


mandli:
What about raising a `ValueError` and then catching it in your case?

mosaiks:
@mandli OK changed to just raising a ValueError

bolliger32:
@mandli OK changed to just raising a ValueError

PR #405, opened 2019-08-22
improving timing with 8 byte integers

rjleveque:
See https://github.com/clawpack/amrclaw/pull/250 for discussion of the changes in this PR.

The `chile2010` example also has a modified `setrun.py` to plot timing stats, using https://github.com/clawpack/visclaw/pull/247

### DIFF
--- a/examples/tsunami/chile2010/setplot.py
+++ b/examples/tsunami/chile2010/setplot.py
@@ -160,6 +160,29 @@ def setplot(plotdata=None):
     plotaxes.afteraxes = add_zeroline
 
 
+    #-----------------------------------------
+    # Plots of timing (CPU and wall time):
+
+    def make_timing_plots(plotdata):
+        from clawpack.visclaw import plot_timing_stats
+        import os,sys
+        try:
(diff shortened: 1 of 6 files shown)


examples/tsunami/chile2010/setplot.py
```diff
@@ -160,6 +160,29 @@ def add_zeroline(current_data):
     plotaxes.afteraxes = add_zeroline
 
 
+    #-----------------------------------------
+    # Plots of timing (CPU and wall time):
+
+    def make_timing_plots(plotdata):
+        from clawpack.visclaw import plot_timing_stats
+        import os,sys
```
mandli: Probably can move this to the top of the module as it should not be an issue importing these.

rjleveque:
I think this is failing because I had to add another variable `tick_cpu_start` to `amr_module.f90`, which is in the amrclaw repository, so we will need to update clawpack/clawpack after https://github.com/clawpack/amrclaw/pull/250 is merged.

rjleveque:
Now passing.

PR #407, opened 2019-10-16
flexibility in netcdf module

bolliger32:
- allow for arbitrary order of x and y dimensions (rather than assuming x is 1st and y is 2nd)
- allow for elevation variable to be indexed (lat, lon) (in addition to (lon, lat))
- add tests to capture this situation
- fix netcdf test so that it wouldn't think it passed if netcdf didn't build. it now "skips" instead


mandli:
Looks like a permission error?

bolliger32:
@mandli oops forgot to remove a test file i was saving to my home directory... works now

mandli:
I figured it was something like that.

PR #409, opened 2019-10-26
update netcdf test and example

bolliger32:
- fix the netcdf_topo test to have reasonable ordering of dimensions in the elevation variable of the netcdf but to swap the order that the dimension IDs get assigned (lat 1, lon 2) to better test new netcdf functionality that can handle this
- add a bowl-slosh-netcdf example that mirrors the bowl-slosh example


rjleveque:
Thanks, works fine!

PR #414, opened 2019-11-08
Wind updates

### LINKED ISSUE #413: Vmax in holland wind profile
@mandli should the max wind speed used in the holland1980 wind profile be using the modified Vmax (adjusting to the top of the boundary layer and removing translational speed) instead of the raw Vmax?

i.e. in lines 470-472 of `model_storm_module.f90`, should `mws` be `mws_mod`?
```
                ! Speed of wind at this point
                wind = sqrt((mwr / r)**B &
                        * exp(1.d0 - (mwr / r)**B) * mws**2.d0 &
                        + (r * f)**2.d0 / 4.d0) - r * f / 2.d0
```

bolliger32:
Fixes #413. Also, slightly rearranges the order of some operations used to add translation speed back to the holland wind profile, to bring in line with ADCIRC.

Also, updated the wind drag limit in the Garret model to match that which is now used in ADCIRC's Garret model.

Also, updated the `storm_surge` test so that it actually uses winds forced from the Ike ATCF track (and, in doing so, adds a regression test that incorporates wind drag)


bolliger32:
OK @mandli this should be good to go, including a new test. The question is whether you want to adjust the upper limit for the Garratt drag to match the change that ADCIRC implemented at some point. That's what I've currently done, but we could just as easily take that out or, slightly less easily, make it a runtime parameter.

src/2d/shallow/surge/model_storm_module.f90
```diff
@@ -407,7 +407,7 @@ subroutine set_holland_1980_fields(maux, mbc, mx, my, xlower, ylower,    &
         ! Local storage
         real(kind=8) :: x, y, r, theta, sloc(2), B
         real(kind=8) :: f, mwr, mws, Pc, Pa, dp, wind, tv(2), radius
-        real(kind=8) :: mod_mws, trans_speed, ramp
+        real(kind=8) :: mod_mws, trans_speed, ramp, TransSpdX, TransSpdY
```
mandli: Do we need `trans_speed` anymore?  Also can you switch to the python style for variables rather than camelcase?

mandli:
Other than my minor comment I think we can merge this.

src/2d/shallow/surge/model_storm_module.f90
bolliger32: fixed!

bolliger32:
@mandli thanks for taking a look. should be ready if tests pass

PR #416, opened 2019-11-09
Typo in CLE attribution

delgadom:
Should be Chavas, Lin, and Emanuel, not Emmanuel :)
https://journals.ametsoc.org/doi/full/10.1175/JAS-D-15-0014.1


mandli:
I think I do that all the time.

PR #418, opened 2019-11-23
make netcdf build on travis

bolliger32:
We were having some issues with netcdf in our build of clawpack and I realized that we have actually never been testing netcdf on travis. Previously, the netcdf tests would pass if netcdf-based builds failed. At some point, I changed this behavior to "Skip" but since that point, all travis builds have skipped. I installed the appropriate netcdf libraries and now that test is passing "for real".

I also bumped the python 3.x version to 3.7. We could test on 3.5 (and any other 3.x) as well if you think its useful?

Finally, I just added the macOS .DS_Store file to the gitignore since it was annoying me.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ __pycache__
 
 # ipython
 .ipynb_checkpoints
+
+# macOS
+*.DS_Store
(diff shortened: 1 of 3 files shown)


rjleveque:
This seems fine to me, thanks!

Issue #421, opened 2019-12-26
Integer division in multilayer


delgadom:
[`shallow/multilayer/multilayer_module.f90:vel_diff_eigen`](https://github.com/clawpack/geoclaw/blob/master/src/2d/shallow/multilayer/multilayer_module.f90#L214) has a couple instances of integer division that may be a bug:

* [line 235](https://github.com/clawpack/geoclaw/blob/9af7d8200b5ad2f757de0121f4e3e2cf607cae33/src/2d/shallow/multilayer/multilayer_module.f90#L235): `s(1) = - sqrt(grav*total_depth_l) + 0.5d0 * mult_depth_l/ total_depth_l**(3/2) * one_minus_r`
* [line 239](https://github.com/clawpack/geoclaw/blob/9af7d8200b5ad2f757de0121f4e3e2cf607cae33/src/2d/shallow/multilayer/multilayer_module.f90#L239): `s(6) = sqrt(grav*total_depth_r) - 0.5d0 * mult_depth_r / total_depth_r**(3/2) * one_minus_r`

Running with debug flags `-Wall` suggests these exponents are bing coerced to 1:

```
/home/jovyan/git/rhg/coastal/core/clawpack/geoclaw/src/2d/shallow/multilayer/multilayer_module.f90:235:83:

         s(1) = - sqrt(grav*total_depth_l) + 0.5d0 * mult_depth_l/ total_depth_l**(3/2) * one_minus_r
                                                                                   1
Warning: Integer division truncated to constant '1' at (1) [-Winteger-division]
/home/jovyan/git/rhg/coastal/core/clawpack/geoclaw/src/2d/shallow/multilayer/multilayer_module.f90:239:82:

         s(6) = sqrt(grav*total_depth_r) - 0.5d0 * mult_depth_r / total_depth_r**(3/2) * one_minus_r
                                                                                  1
Warning: Integer division truncated to constant '1' at (1) [-Winteger-division]
```

Should be a very easy fix `(3.0/2)` - will file a PR if this looks like a legitimate bug.

mandli:
The velocity difference eigenspace approximation is known to have these issues whereas some of the other methods does not run into this.

delgadom:
I'm referring narrowly to the integer division occurring at `(3/2)`. If I'm not mistaken this results in the integer `(1)` rather than the float `(1.5)`. Not attempting to address anything else about the method :)

If it's not a bug that's fine it just looked like one to me

[...]

PR #422, opened 2019-12-26
Bug fix integer-->float division in vel_diff_eigen

delgadom:
Addresses Issue #421 

Compiling with debug flags suggests these exponents had been coerced to 1


mandli:
This would not correct the division by zero I don't think (although the coercion is a bug).

delgadom:
I don't think there's a divide by 0 error, just accidentally raising to the 1 instead of the 1.5? Could be wrong though.

PR #423, opened 2020-01-02
Lagrangian gauges

rjleveque:
This allows some gauges to be lagrangian to represent massless tracer particles carried with the flow. In each time step (or whenever gauges are updated) the gauge location of these gauges is adjusted using the velocity at the gauge (with a forward Euler update).  

The first pass at a documentation page can be viewed at http://depts.washington.edu/clawpack/sampledocs/lagrangian_gauges.html

This includes a new example in `examples/tsunami/island-particles` that can be viewed at
  http://depts.washington.edu/clawpack/samples/island-particles/README.html

and also a new `tests/particles`.

This won't work without corresponding changes in the repositories amrclaw, pyclaw, and visclaw.  I will do PRs on each of those shortly.

It should also be possible to get working versions of all repositories via:

```
    git clone http://github.com/rjleveque/clawpack clawpack_lagrangian_gauges
    cd clawpack_lagrangian_gauges
    git submodule init
    git submodule update
```
and then set `$CLAW` to point to this version,
and also `$PYTHONPATH`, or else use `pip install -e .`


rjleveque:
Passing after merging https://github.com/clawpack/clawpack/pull/176

PR #428, opened 2020-02-10
Fgmax improvements

rjleveque:
This PR introduces some new features and naming conventions and is not backward compatible.

The fgmax monitoring has been sped up considerably by completely refactoring most of the routines `src/2d/shallow/fgmax*.f90` and also some changes to `tick.f`.  The old code looped over all fgmax points many times for each update time and was extremely slow with many fgmax points. The new version has been pretty extensively tested on problems with several million fgmax points.

A new method of specifying fgmax points has been introduced (in addition to the ones already allowed) and is supported in the new version of `fgmax_tools.py` and the Fortran code. If `fg` is of class `fgmax_tools.FGMaxGrid` and has `fg.point_style==4` then `fg.xy_file` is a path to a file that has the same format as topofile DEMs with `topo_type=3`, but the `Z` value is 1 for points to be used as fgmax points, 0 otherwise.  

Setting up fgmax grids is now done mostly in `setrun.py` and instead of creating a file `fgmax.data` doing `make data` instead creates a file `fgmax_grids.data` that has more information in it.

A new example `examples/tsunami/radial-ocean-island-fgmax` has been added that illustrates this.  See http://depts.washington.edu/clawpack/samples/radial-ocean-island-fgmax/README.html for sample output.

I still need to write a test version of this to add to `tests`.

Note that much of the development of these changes was done in other repositories, for some of the recent work see the commits in https://github.com/rjleveque/new_features_for_v5.7.0/tree/fgmax_improvements


mandli:
It looks like the `fgmax_grids.data` is not passing yet.

rjleveque:
Yes, it turns out there was an fgmax test and it's not passing.  Also some other examples needed updating.  I'm working on it.

mandli:
Is this ready to go?

rjleveque:
Not yet, still sorting a few things out...

rjleveque:
@mandli, I think this is ok to merge now if it looks ok to you.

PR #429, opened 2020-02-12
Update with additional wind models 

hudaquresh:
This is an update to geoclaw fortran source code. The following six parameterized wind models are added to the storm module: holland10, rankine, modified-rankine, SLOSH, DeMaria, and the Chavas-Lin-Emanuel (CLE) model. 

The CLE is not working properly. For now an implementation error is raised if the CLE model is used; however, the model is included in the fortran code.


mandli:
I think you need to remove some output files from this and probably rebase on the newest version of GeoClaw, which should take care of some of these commits.

PR #431, opened 2020-02-17
Add additional wind models

hudaquresh:


src/2d/shallow/surge/model_storm_module.f90
```diff
@@ -540,15 +674,1069 @@ subroutine set_CLE_fields(maux, mbc, mx, my, xlower, ylower,    &
...
+        !First, interpolate all of the relevant tracked storm parameters.
+        call get_cle_storm_data(t,storm,sloc,tv,mwr,mws,Pc,radius)
+
+        ! Other quantities of interest 
+        Pa = ambient_pressure 
+
+        ! Additional quantities of interest 
+        i = storm_index(t, storm)
+        f = coriolis(sloc(2))
+        tv = storm%velocity(:, i)
```
bolliger32: @hudaquresh @mandli just making sure I understand whats going on here b/c we had previously included the CLE model but had the assignments to `tv` through `mwr` commented out. What's the rationale behind overwriting the values obtained in `get_cle_storm_data` above with new values here? 

Also I think `Pa` is defined twice.

mandli: @hudaquresh I think this follows the previous Holland model?

PR #433, opened 2020-02-20
Fix annoying numpy issue

bolliger32:
With some versions of numpy, it won't accept a float64 dtype as the `num` kwarg for `np.linspace`. I noticed this b/c it failed the build of our ClimateImpactLab/clawpack fork when we merged in the latest changes to clawpack/clawpack


mandli:
I have noticed this from time to time as well with Python 3.

PR #434, opened 2020-02-22
Multilayer gauge fix

rjleveque:
Lagrangian gauges aren't yet handled in multilayer code but some variables need to be added to the `gauge_type`. 

There aren't used unless doing a restart, but the latest gfortran caught this and `tests/multilayer` didn't pass.

I also decided to check that restarting works and it didn't because of another issue: In `amr2.f90` it is now necessary to call `set_geo` and `set_multilayer` before `set_gauges` so the number of layers is set.

At some point, need to investigate whether other `set_*` routines should best be called before or after `restrt`, depending on whether the user should be allowed to modify various things in the restart from the original run.

But I think things are working again as in v5.6.1.

### DIFF
--- a/src/2d/shallow/amr2.f90
+++ b/src/2d/shallow/amr2.f90
@@ -464,6 +464,11 @@ program amr2
 
         ! moved upt before restrt or won't properly initialize 
         call set_fgmax()   
+        
+        ! need these before set_gauges so num_out_vars set right for gauges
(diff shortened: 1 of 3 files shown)


mandli:
I guess the *.data files could change between restarts huh?  Maybe we should raise an issue about that?

rjleveque:
@mandli, does this look ok to merge?

mandli:
Looks good to me!

PR #436, opened 2020-02-24
Correctly handle zero coordinate values in read_atcf

tovogt:
In some ATCF files, such as https://ftp.nhc.noaa.gov/atcf/archive/2001/aal032001.dat.gz, zero coordinate values are sometimes stored as "0" instead of "0N" so that the previous parsing routine raised a ValueError ("could not convert string to float:").

### DIFF
--- a/src/python/geoclaw/surge/storm.py
+++ b/src/python/geoclaw/surge/storm.py
@@ -357,11 +357,15 @@ class Storm(object):
             self.classification[i] = data[10]
 
             # Parse eye location - longitude/latitude order
-            if data[6][-1] == "N":
+            if data[6] == "0":
+                self.eye_location[i, 1] = 0.0
+            elif data[6][-1] == "N":
                 self.eye_location[i, 1] = float(data[6][0:-1]) / 10.0
             else:
                 self.eye_location[i, 1] = -float(data[6][0:-1]) / 10.0
-            if data[7][-1] == "E":
+            if data[7] == "0":
+                self.eye_location[i, 0] = 0.0
+            elif data[7][-1] == "E":
                 self.eye_location[i, 0] = float(data[7][0:-1]) / 10.0
             else:
                 self.eye_location[i, 0] = -float(data[7][0:-1]) / 10.0


tovogt:
Ah, wait. Maybe we should consider handling those lines in a different way. The affected line in the linked file seems to contain corrupt data:

```
...
AL, 03, 2001080300, 03,  NGM,   0, 263N,  854W,  13, 1016, XX,  34, NEQ,    0,    0,    0,    0, 
AL, 03, 2001080300, 03,  NGM,  12, 276N,  867W,  14, 1014, XX,  34, NEQ,    0,    0,    0,    0, 
AL, 03, 2001080300, 03,  NGX,   0,   0 ,    0 ,   0,    0, XX,  34, NEQ,    0,    0,    0,    0, 
AL, 03, 2001080300, 03,  UKM,   0, 263N,  856W,   0,    0,   ,   0,    ,    0,    0,    0,    0, 
AL, 03, 2001080300, 03,  UKM,  12, 269N,  869W,   0,    0,   ,   0,    ,    0,    0,    0,    0, 
AL, 03, 2001080300, 03,  UKM,  24, 271N,  871W,   0,    0,   ,   0,    ,    0,    0,    0,    0, 
...
``` 

So, maybe we should rather discard such lines completely in an earlier step. We could discard every line where columns 6-9 equal "0". What do you think?

mandli:
Usually when we have duplicated lines like this we only use the first one but in this case it looks like the first one also has issues possibly.  Do you know off hand what column 5 means?

tovogt:
That's the forecast period (in hours). 0 means it's best track information. Documentation of ATCF best track files can be found here: https://www.nrlmry.navy.mil/atcf_web/docs/database/new/abdeck.txt

mandli:
Oh, meant the column before that.  In any case we have not yet really tackled how to deal with forecast data.  Thoughts?

tovogt:
Unfortunately, I couldn't find a complete list of acronyms used in that column. NGM seems to be "nested grid model", NGX is probably "Navy Operational Global Atmospheric Prediction System (NOGAPS)". UKM and UKX are most probable forecasting models by the UK Met Office.

Forecast data (especially pre-2004) often has wind radius information when its missing in the corresponding best track data. That's why I would like to use it as a data source in `max_wind_radius_fill` and `storm_radius_fill`.

The date for each row only gives the date of generation while the data in those rows correspond to the state of the storm at the forecasted time. An entry with the timestamp "2001080300" (midnight of August 3, 2001) and the forecasting period "12" will give information about the storm's state at noon on August 3, 2001.

I see only now that the current parsing logic will misattribute data since it ignores the forecast period. Furthermore, by considering only the first entry for duplicates it will sometimes discard valuable information that is only available in the later entries (such as wind radius).

I created a new pull request (#437) to demonstrate what I would expect from an ATCF parser that takes forecasting into consideration.

mandli:
I think #437 makes sense to me.  Someone could always sub-class the storm object to get the behavior they want and unless someone has an idea what to use that seems like the best path forward.

tovogt:
Cancelling in favor of #437

PR #437, opened 2020-02-25
Consider forecasting data in Storm.read_atcf

tovogt:
The date of each entry is recalculated using the forecast period. In case of duplicates, it will choose the best (in terms of forecasting period) available data. Requires pandas.

See #436 for previous discussion.


mandli:
There are other places in `storm.py` that we depend on pandas.  I would suggest that you `import` it in the function just to limit it's use and skip the test of `pandas` is not present.

mandli:
Hi @tovogt, what's the status of this PR?

tovogt:
Just yesterday I made a rebase to the base branch. A check for pandas is in the `read_atcf` function call and in the storm file IO test. All tests are passing. Seems like it's ready to be merged :)

PR #440, opened 2020-03-10
Updates to support set_eta_init and force_dry options

rjleveque:
This set of changes supports two improvements that both affect the same routines and are somewhat intertwined although they address different issues:

1. `set_eta_init.f90` is a new library routine that can be called to set the initial `eta = h+B` elevation in some way other than simply using a single value `sea_level` everywhere.  In particular this is needed if there are regions that subside (or uplift) during an earthquake and a fine grid is not initialized until some later time. See [this notebook](http://www.clawpack.org/new_features_for_v5.7.0/notebooks_html/SetEtaInit.html) for some more information.   It can also be used to initialize an inland lake with a different surface elevation than the ocean, for example.

2. `force_dry_init` can be used to identify some regions that should be initialized to dry land even if the topography is below sea level.  This is important to properly handle dry land behind dikes or levies in some tsunami or storm surge simulations.  See [this notebook](http://www.clawpack.org/new_features_for_v5.7.0/notebooks_html/ForceDry.html) for some examples of how to use this, and how to pre-process a topofile to create a `force_dry_init` array using the [marching_front  module](https://github.com/clawpack/geoclaw/blob/master/src/python/geoclaw/marching_front.py)  recently added to GeoClaw.

I'm still working on testing and developing some better examples, but would welcome feedback at this point.


rjleveque:
Does anyone know why travis no longer seems to run automatically on Clawpack PRs? And I don't even see how to start it, so something must have changed in the way things have to be configured?  I can't seem to find travis under `settings` any more.

rjleveque:
I have added some new documentation on these features and also a new example in `examples/tsunami/eta_init_force_dry`.  

In this example I'm also experimenting with using Jupyter notebooks to make the required input files and to run a set of geoclaw tests and discuss the results of each.  See the notebook links at
[README.html](http://depts.washington.edu/clawpack/sampledocs/eta_init_force_dry/README.html)
Feedback welcome!

Please also check out the new documentation:
- [set_eta_init](http://depts.washington.edu/clawpack/sampledocs/dev_v5.7.0/set_eta_init.html)
- [marching_front](http://depts.washington.edu/clawpack/sampledocs/dev_v5.7.0/marching_front.html)
- [force_dry](http://depts.washington.edu/clawpack/sampledocs/dev_v5.7.0/force_dry.html)

rjleveque:
It looks like travis did run but didn't get linked from this PR as expected.
  https://travis-ci.org/github/clawpack/geoclaw/builds/671339639

I found it via travis-ci.org, and a bunch of things are failing with Python3, but strangely not with Python2 (and it works fine with 3 on my laptop),

mandli:
I just saw someone complaining about this as well.  Maybe the old version of the travis scripts have been phased out?

rjleveque:
I switched to Python 3.6.10 rather than 3.7 in `.travis.yml` and now it is passing.

rjleveque:
I think this is ready to merge unless there are any concerns or suggestions.

mandli:
There are two files being added that I think may need some modification:
1. `make_input_files.html` has the twitter bootstrap CS and javascript code in it.
1. `make_intput_files.ipynb` has results that have been run in it (e.g. images).

I think that `run_geoclaw.html` and `run_geoclaw.ipynb` might also have the same issue.  I am assuming that the html is from the notebooks?

rjleveque:
Yes, I forgot to run nbstripout.  Thanks for catching that @mandli.

Regarding the twitter bootstrap stuff in the .html files, that is generated automatically by nbconvert applied to the .ipynb files.   

I thought it was good to include these so that people can read the notebook and see the output it generates without having to run the notebook, since many Clawpack users may not be notebook users.  But they are rather big, particularly `run_geoclaw.html`, and would be even bigger if animations were included like in some of the ones in the apps repository.  So maybe this isn't a good idea and the notebook output should just be stored in the gallery?

rjleveque:
I removed the .html files and instead added a link to the gallery, to this page:
http://www.clawpack.org/gallery/_static/geoclaw/examples/tsunami/eta_init_force_dry/README.html

mandli:
I figured that's why you wanted to include them.  It's a bit annoying that bootstrap stuff automatically.  I think there may be a way to simply link to that but I forget how to set that up.  If this looks good to everyone else I think this can be merged.

PR #442, opened 2020-04-08
again fix find to where in dtopotools

rjleveque:
Because I did it incorrectly in #427.


rjleveque:
@mandli: travis tests are at least showing up again, on this PR (but failing still)

rjleveque:
Travis tests failing as in #440, but pass locally, so some issues still to be addressed with Python versions.  Merging this minor change.

PR #449, opened 2020-05-20
Add error handling for atcf files with fewer than 10 columns

### LINKED ISSUE #448: read_atcf IndexError
I'm getting an `IndexError` when calling `surge.storm.Storm` on this atcf file ([bep081994.dat](https://ftp.nhc.noaa.gov/atcf/archive/1994/)). The issue is the first 10 or so lines of the file have fewer columns than expected by the code.

mayaanorman:
Fixes #448.

### DIFF
--- a/src/python/geoclaw/surge/storm.py
+++ b/src/python/geoclaw/surge/storm.py
@@ -354,7 +354,10 @@ class Storm(object):
             #             self.time_offset = self.t[i]
 
             # Classification, note that this is not the category of the storm
-            self.classification[i] = data[10]
+            try:
+                self.classification[i] = data[10]
+            except IndexError:
+                self.classification[i] = ''
 
             # Parse eye location - longitude/latitude order
             if data[6][-1] == "N":
@@ -373,7 +376,7 @@ class Storm(object):
                 self.max_wind_speed[i] = -1
             try:
                 self.central_pressure[i] = units.convert(float(data[9]), 'mbar', 'Pa')
-            except ValueError:
+            except (ValueError, IndexError):
                 self.central_pressure[i] = -1
 
             # if this is a shortened line - does not contain max wind radius


src/python/geoclaw/surge/storm.py
```diff
@@ -354,7 +354,10 @@ def read_atcf(self, path, verbose=False):
             #             self.time_offset = self.t[i]
 
             # Classification, note that this is not the category of the storm
-            self.classification[i] = data[10]
+            try:
+                self.classification[i] = data[10]
+            except IndexError:
+                self.classification[i] ''
```
mandli, line 360: Missing an equal sign here.

mandli:
Instead of doing two try/except blocks here I might suggest checking to see the length of the array directly and handling that case.  Beyond that I have two questions:
 - It looks like the central pressure is not in this data file at all, we should probably handle that as well.
 - Does the parser work on this at all?  I would guess that it would get confused when the length of the rows changes.

src/python/geoclaw/surge/storm.py
mayaanorman: Thanks! My fix for that typo clearly didn't make it into my commit.

mayaanorman:
@mandli Okay, I can do that -- handle differing line lengths directly rather than with try and except statements. 

(1) _It looks like the central pressure is not in this data file at all, we should probably handle that as well._ 

Currently, missing values for max wind speed, central pressure, and storm radius gets filled with -1. When the code attempts to convert a missing value to a float the result is a value error. Here's a code block to hopefully better contextualize what I'm saying.

```
try:
    self.max_wind_speed[i] = units.convert(float(data[8]), 'knots', 'm/s')
except ValueError:
    self.max_wind_speed[i] = -1
```
Do you want missing data to be handled differently?

(2) _Does the parser work on this at all? I would guess that it would get confused when the length of the rows changes._

The parser does work. The try and except statements in the code seem to handle the varying row lengths well. Although ATCF data line lengths seem to vary a fair amount from file to file and even within some files, I have yet to find any files where the data is misplaced. Based on the (documentation)[https://www.nrlmry.navy.mil/atcf_web/docs/database/new/abdeck.txt], there should always be 37 columns in each data line. This is definitely not the case. I've found one case where there are more than 37 columns in a data line and many cases where there are fewer. I have yet to find any examples where the data has been placed in the wrong column though — meaning if the column is present its index always corresponds to the same value.

mandli:
wrt the value of `-1` it really should be a `nan` but I am not sure the rest of the code is set up to handle those yet.  Let's not do that here but I will raise an issue.

mayaanorman:
@mandli ah okay, great. Do you still want me to replace the try and except statements with code that directly handles differing line lengths? Or are you happy enough with the current set up now?

mandli:
Yes if you get a chance but when you have time.

mayaanorman:
Great, will do. Thanks @mandli.

PR #455, opened 2020-06-15
in read_netcdf, allow variables x,y in place of lon,lat

rjleveque:
Some netCDF files use `x` and `y` as the horizontal variables rather than `lon`, `lat`, so check for this case instead of failing.

We could add `xvar` and `yvar` input parameters similar to `zvar`, but that would mean a new call signature and probably there aren't other variable names being used.

It now also prints out `f.variables` if it doesn't find one of the expected names.

### DIFF
--- a/src/python/geoclaw/topotools.py
+++ b/src/python/geoclaw/topotools.py
@@ -1747,8 +1747,21 @@ def read_netcdf(path, zvar=None, extent='all', coarsen=1, return_topo=True,
     else: 
         f = netCDF4.Dataset(path, 'r')
 
-    x = f.variables['lon']
-    y = f.variables['lat']
+    if 'lon' in f.variables:
+        x = f.variables['lon']
+    elif 'x' in f.variables:
+        x = f.variables['x']
+    else:
+        print('*** f.variables = ',f.variables)
+        raise ValueError("*** Unrecognized x, lon in netCDF file")
+
+    if 'lat' in f.variables:
+        y = f.variables['lat']
+    elif 'y' in f.variables:
+        y = f.variables['y']
+    else:
+        print('*** f.variables = ',f.variables)
+        raise ValueError("*** Unrecognized y, lat in netCDF file")
 
     # for selecting subset based on extent, convert to arrays if netCDF4 used:
     #if not return_xarray:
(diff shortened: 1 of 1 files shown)


mandli:
I was working through a NetCDF topo problem today with someone and I ran into this problem.  I thought at some point we had a dictionary that did this but now I do not see it.  The dictionary is nice as you can check to see if each key/value pair is available and use a default otherwise.  I can code up what I was thinking but it basically is
```python
var_mapping = {} # <-- input
x_var = var_mapping.get('x_var', 'lon')
y_var = var_mapping.get('y_var', 'lat')
if not(x_var in f.variables.keys() and y_var in f.variables.keys()):
    raise ValueError('   ')
```
alternatively we could provide a default dictionary argument but that gets clunky when writing out the function signature.  By default this really should be an empty variable though.

rjleveque:
It would still be nice to have it try the standard possibilities first, e.g. both `lon` and `x` before expecting the user to provide something.  But if we do want to allow the user to provide names for all three (x,y,z) then maybe a dictionary is nicer than three separate parameters.  Not sure which is simpler.

Do you think there are variable names other than `x,y` or `lon,lat` that might be used for the horizontal?  I suppose maybe `longitude, latitude`?

mandli:
The above code defaults to `'lon'` and `'lat'` if no keys are present for `x_var` and `y_var` or is that not what you meant?

In terms of other defaults I have seen a few but by far those two are the most widely used with the exception of non-lat/long coordinate systems of course.

rjleveque:
The way I wrote it, first it tries 'lon' and if that doesn't work it tries 'x'.  We might want to first try both these choices before expecting the user to provide something different.  

(Similarly, the current code checks for 'z', 'elevation' and 'Band1' before complaining if `zvar` is not passed in.)

The travis tests passed if you click on Details, so I don't know why this PR still says they are in progress.

rjleveque:
I suggest for now we merge in this PR since it doesn't change the calling sequence and is backward compatible, and then think more about redoing how x,y, and also z are handled in some future major release.

mandli:
Adding `var_mapping` would not change the calling sequence if it is a key word argument though.  We could also program around the existing `z_var` issue.  That seems like an approach that would add the functionality but also allow for backwards compatibility?

mandli:
We are going to merge this in and then raise an issue (#459) that we have duplicated functionality in topotools right now as the `read_netcdf` function exists as well as `read` in the `Topography` class.

mandli:
Forgot that I needed to push a button :P

PR #456, opened 2020-06-15
WIP: initialize slope=0 for each var in filpatch

rjleveque:
Minmod limiting is used to set slopes for interpolating from coarser grids when needed in patch ghost cells.  In case `up_slope*down_slope < 0` the value of `slope` is left unchanged, so it should be initialized to 0.  Currently this is done before the loop on components `n` and should be done again for each momentum component or potentially the slope from a previous variable will be used.

This should affect things only in ghost cells where there is a local max or min in one of the variables, but conceivably is related to instabilities that have sometimes been observed.

Since this can affect results, I would like to test some more before merging this in.


rjleveque:
Note: This modification causes slight changes in regression data for `tests/storm_surge` and `tests/chile2010_adjoint`.

rjleveque:
I updated this PR by merging in other recent changes to master, including #478, and so I expected the regression data might need to be modified yet again.   But it doesn't because it turns out this PR includes the same fix as #478, i.e. commits 1626bca and 7a4397c did the same thing.

Bad memory on my part...

rjleveque:
After more extensive testing of this together with #478 this seems ok and is a necessary bug fix.

PR #461, opened 2020-06-29
fix fgmax_frompatch index error

rjleveque:
This only showed up in a case where a patch edge happened to be very close to the edge of the fgmax grid.


rjleveque:
Not sure why travis is failing, I'm also getting errors on my laptop now and will investigate.

But there's also another related index issue that just came up so I will probably add to this PR, so don't merge yet.

rjleveque:
I think this is ok to merge.

PR #464, opened 2020-08-01
fix memory and cleanup netcdf mod

bolliger32:
The main fix in this PR is to make the `var_ids` vector in the `netcdf_module` allocatable. Previously, it had been a static array of length 2, but all netcdf topo files will have at least 3 vars (x, y, z), which was the length of the array that was returned from the `nf90_inq_varids` call. So I'm not sure how this was working for so long. I switched to an allocatable array, which gets allocated after reading the number of variables in the file in an earlier call. Somehow, this memory issue had not been causing segfaults unless I had the `fstack-protector` flag on, in which case it caused a segfault. But even when it was not segfaulting, it was giving wacky behavior for runs with netcdf topo. In particular, it was giving different results depending on whether you redirected stdout and stderr to files or not. I'm not sure of the context in which this would have affected previous results, but once I fixed this I'm no longer getting different results depending on output redirection.

Other than that, I just cleaned up the netcdf_topo module a bit, getting rid of unused variables. I only did this b/c in my debugging I just started following my way up the compiler warnings and this is what I got to.

I originally assumed that I had introduced this bug with #400 but from the looks of that PR it had been around before that too.

### DIFF
--- a/src/2d/shallow/multilayer/setaux.f90
+++ b/src/2d/shallow/multilayer/setaux.f90
@@ -112,7 +112,7 @@ subroutine setaux(mbc,mx,my,xlow,ylow,dx,dy,maux,aux)
                 xm = xlower + (ilo+ii-1.d0) * dx
                 xp = xlower + (ilo+ii) * dx
                 x = 0.5d0*(xm+xp)
-                aux(1,ii,jj) = test_topo(x, y)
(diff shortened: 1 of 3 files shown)


mandli:
Thanks @bolliger32!

Issue #466, opened 2020-08-11
Some errors in wind field models


bolliger32:
As I was creating #467, I noticed that a couple of the wind field models seem to have errors in them:

- The CLE model calculates the circular windspeed by removing the translational wind speed, but then does not use it anywhere in the model. I'm not sure if CLE is supposed to be using the circular wind speed in the algorithms it implements, but I think so?
- The Willoughby field has this weird clause at the bottom:
```
                ! Speed of wind at this point
                if (r < 0.9d0 * mwr) then 
                    wind = mws * (r / mwr)**0.45d0
                !else if (r > 0.9d0 * mwr & r < 1.1d0 * mwr) then 
                else if (r > 0.9d0) then 
                    wind = 2.d0 
                else 
                    wind = 1.d0 
                end if  
```
Seems like there are some bugs to fix there.

Once these bugs are addressed, there should probably be a regression test for each wind field. I didn't look at any of the other wind fields in depth but just noticed those that I mentioned. Might be worth a detailed look.

mandli:
@hudaquresh can you take a look at this?

PR #467, opened 2020-08-11
Clean up model_storm_module and bound azimuthal wind at 0

### LINKED ISSUE #465: Vortex models can give negative gradient winds (at least using IBTrACS)
I'm not exactly sure how this manifests in geoclaw wind fields but we found in our own wind field modeling (external to geoclaw but using a similar approach) that we were seeing some really high estimated winds for storms that had dissipated over land. Ultimately, this was due to getting a negative number when we subtracted the translational velocity from the observed max wind speed. (i.e. https://github.com/clawpack/geoclaw/blob/master/src/2d/shallow/surge/model_storm_module.f90#L428). We're making sure this has a minimum at 0 going forward, and I think that would probably make sense for geoclaw winds as well. To illustrate this issue, here's Hurricane Frederic 1979.

In IBTrACS, with color corresponding to reported max wind speed (the second figure is just resampled to 30min intervals):
image

Here's our representation of max wind speed experienced in each pixel, using a similar approach to geoclaw:
image

And here's where we clip circular wind speed to be >=0:
image

I'll work on an implementation in the `model_storm_model` file

bolliger32:
The main change from this PR is to close #465 by clipping azimuthal wind speed at 0 (no negative values). In addition, I cleaned up the model_storm_module file a bit by pulling out common pre- and post-processing functionality into separate subroutines (including this bounding at 0 function). I noticed a few errors in some of the wind models that I noted in #466 but didn't address as my knowledge of these wind models is limited

### DIFF
--- a/src/2d/shallow/surge/model_storm_module.f90
+++ b/src/2d/shallow/surge/model_storm_module.f90
@@ -389,13 +389,16 @@ contains
  
     ! 1. Adjust max wind speed by removing translational speed (bounding at 0).
 
-    ! 2. Convert wind speed at 10 m to top of atmospheric boundary layer
+    ! 2. If specified, convert surface winds to PBL winds
     ! ==========================================================================
-    pure subroutine adjust_max_wind(tv, mws, mod_mws)
+
+    pure subroutine adjust_max_wind(tv, mws, mod_mws, convert_height)
 
         real (kind=8), intent(inout) :: tv(2)
         
         real (kind=8), intent(in) :: mws
+
+        logical, intent(in) :: convert_height
 
         real (kind=8), intent(out) :: mod_mws
 
@@ -412,7 +415,7 @@ contains
         end if
 
         ! Convert wind speed (10 m) to top of atmospheric boundary layer
-        mod_mws = mod_mws / atmos_boundary_layer
+        if (convert_height) mod_mws = mod_mws / atmos_boundary_layer
 
     end subroutine adjust_max_wind
 
@@ -425,8 +428,8 @@ contains
     !    This is tapered to zero as the storm wind tapers to zero toward the eye of the 
     !    storm and at long distances from the storm. Avoid divide-by-0 error
 
-    ! 2. Convert wind velocity from top of atmospheric boundary layer to wind
-    !    velocity at 10 m above the earth's surface.
+    ! 2. If specified, convert wind velocity from top of atmospheric boundary layer to 
+    !    wind velocity at 10 m above the earth's surface.
 
     ! 3. Convert from 1 minute averaged winds to 10 minute averaged winds
 
@@ -435,13 +438,16 @@ contains
     ! 5. Apply distance ramp to limit scope
     ! ==========================================================================
     pure subroutine post_process_wind_estimate(maux, mbc, mx, my, i, j, wind, aux, &
-        wind_index, pressure_index, r, radius, tv, mod_mws, theta, Pa)
+        wind_index, pressure_index, r, radius, tv, mod_mws, theta, Pa, &
+        convert_height)
 
         integer, intent(in) :: maux, mbc, mx, my, i, j
         real(kind=8), intent(inout) :: aux(maux,1-mbc:mx+mbc,1-mbc:my+mbc), wind
 
         real (kind=8), intent(in) :: tv(2), mod_mws, theta, Pa, r, radius
         integer, intent(in) :: wind_index, pressure_index
+
+        logical, intent(in) :: convert_height
 
         real (kind=8) :: trans_speed_x, trans_speed_y, ramp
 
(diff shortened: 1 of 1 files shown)


src/2d/shallow/surge/model_storm_module.f90
```diff
@@ -383,6 +383,92 @@ subroutine get_storm_data(t, storm, location, velocity, max_wind_radius,  &
...
 
+    ! ==========================================================================
+    ! A set of common functions applied before calculating the circular component of 
+    ! wind speed using a particular wind model.
+ 
+    ! 1. Adjust max wind speed by removing translational speed (bounding at 0).
+
+    ! 2. Convert wind speed at 10 m to top of atmospheric boundary layer
+    ! ==========================================================================
+    subroutine adjust_max_wind(tv, mws, mod_mws)
```
mandli, line 394: Maybe make this a `pure function` so that it can be inlined automatically?

src/2d/shallow/surge/model_storm_module.f90
```diff
@@ -383,6 +383,92 @@ subroutine get_storm_data(t, storm, location, velocity, max_wind_radius,  &
...
+    ! 2. Convert wind velocity from top of atmospheric boundary layer to wind
+    !    velocity at 10 m above the earth's surface.
+
+    ! 3. Convert from 1 minute averaged winds to 10 minute averaged winds
+
+    ! 4. Add back in translational velocity
+
+    ! 5. Apply distance ramp to limit scope
+    ! ==========================================================================
+    subroutine post_process_wind_estimate(maux, mbc, mx, my, i, j, wind, aux, &
```
mandli, line 435: Again, is it possible to do with a `pure` function?

mjberger:
coming in late to this discussions. But I"ve had pure functions not work in openmp with multiple threads. Do you understand this?

In my case it was an "iadd" pure function, and I had to put it back to being a statement function.

 — Marsha

mandli:
Nope, not sure why that would happen.  Do you know what the situation was?

mjberger:
Nope. But it completely surprised me. I had to track it down with print statements.

 — Marsha

mandli:
Ug, Fortran

bolliger32:
@mandli @mjberger I'll give pure functions a try... with the caveat that I don't really know what I'm doing in fortran :) 

I wonder if it's possible to make the ike test use openmp. Not that it would catch every potential gotcha case, but might be one line of defense.

src/2d/shallow/surge/model_storm_module.f90
bolliger32: OK i added the pure modifier. Was that what you meant? Or is there something deeper. I wasn't able to figure out much about pure procedures online so I'm not really sure what I'm doing. I dont' think this can easily be made into a function? B/c it's modifying tv in addition to producing mod_mws

src/2d/shallow/surge/model_storm_module.f90
bolliger32: Same thing here. I just slapped pure on there, since changing the subroutine to a function would be tricker since it modifies a couple of values.

bolliger32:
And just to confirm i did run the ike test on my machine with openMP and it didn't seem to cause any issues.

bolliger32:
OK I went down a bit of a rabbit hole into these vortex models in trying to clean this up. I've updated a few more things:

General cleaning
-----------------
- I pulled out the Holland B parameter calculation into its own pure function
- I pulled out the polar coordinate calculation into its own pure function
- I pulled out the pressure difference calculation into its own pure function (which for some models was bounding at 100Pa but not for all models)
- I added some references for the models in the docstrings

Model updating
----------------
- I updated several models where the full max observed wind was being used rather than the max of the circular wind speed
- I uncommented a line for calculating "x" in the Holland 2010 model, so it should now work
- I did NOT touch the CLE model as this is above my pay grade :) 
- I corrected Holland 2010 to use surface level obs (while all the others used the gradient level winds)
- I updated the SLOSH model to use the formula for combining translational speed and circular speed that is actually employed in SLOSH
- I provided an updated estimation of alpha in the modified rankine model based on a 2005 paper
- I finished up what looked like a partially finished Willoughby model

These are a lot of changes, so I'm not offended if you want me to chuck some of them. It's probably worth @hudaquresh taking a look, since I'm sure you're pretty familiar with whats going on here and can let me know if/where i screwed up :) 

These changes should additionally close #466

src/2d/shallow/surge/model_storm_module.f90
mandli: Did it actually compile then?  I thought `pure` could only be added to a function.

bolliger32: it did...but i'm wondering if it just ignored the pure modifier? Here's what I was using to try to figure out how pure works: http://www.lahey.com/docs/lfpro78help/F95ARPURE.htm. Suggests that it can be applied to subroutines, but seems like these ones don't actually meet the criteria

mandli:
Sorry, been super busy with the beginning of a semester of some sort.  Given that @hudaquresh is studying for an exam these days so I took a look.  All the changes look good to me but it's hard to give a general commentary given we do not have full test coverage on all of these.

In terms of the CLE model though, my guess is that it does not need the ramping function as it actually limits to zero unlike most of the others.  We should open an issue if there isn't one already to figure out the CLE model cleanup.

PR #468, opened 2020-08-19
fgmax_tools: close file pointer after reading

tovogt:
Tiny fix, that removed a `ResourceWarning: unclosed file` for me.


rjleveque:
Works for me.  Thanks for cleaning up the white space too!

PR #469, opened 2020-08-27
Fix fetch_noaa_tide_data() method to match current NOAA API formatting

jkapilian:
Previously, this method would expect different formatting from the NOAA API, leading to a Value Error being thrown.

### DIFF
--- a/src/python/geoclaw/util.py
+++ b/src/python/geoclaw/util.py
@@ -271,7 +271,7 @@ def fetch_noaa_tide_data(station, begin_date, end_date, time_zone='GMT',
 
     # fetch water levels and tide predictions
     date_time, water_level = fetch(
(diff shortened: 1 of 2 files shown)


mandli:
Hi @jkapilian, it looks like the tests are failing.  Can you take a look?

jkapilian:
It seems the test file also was set up to expect the old NOAA API output so it should pass the tests now

mandli:
Did you change the output to the new version then?  It seems like the tests are still failing.

jkapilian:
Yeah I just saw that I changed the output to the new version so I'll have to look and see what's causing it this time

jkapilian:
Looks like I just missed one of the tests in test_util.py it's working now

mandli:
Thanks!

PR #474, opened 2020-09-09
warn instead of fail if etopo1 data can't be read from thredds server

rjleveque:
The NOAA thredds server is sometimes down, making travis fail. 
This would issue a warning instead. 

@mandli: Look ok to do it this way?

### DIFF
--- a/tests/test_etopo1.py
+++ b/tests/test_etopo1.py
@@ -5,6 +5,7 @@ import os
 import numpy
 import nose
 import os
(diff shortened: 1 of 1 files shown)


rjleveque:
Now the thredds server seems to be back up so this passed without the new warnings option.

rjleveque:
I'm going to merge to retest https://github.com/clawpack/clawpack/pull/188

PR #478, opened 2020-09-20
WIP: fix filpatch Y slope sign

rjleveque:
Should be computed based on sign of difference in j, not in i.

@mjberger: does this look like the right fix for the bug you found?

This affects how coarse grid data is interpolated to fine patch boundaries, and it seems like the slope of the linear interpolant could have often been negated, so I'm surprised this hasn't caused more problems in the past.

This change modifies results, and the travis tests will no longer pass until we update those.  I'll take a look at `tests/chile2010_adjoint`.   @mandli:  I think `tests/storm_surge` will also fail, at least it does on my laptop.  Do you want to take a look at this and perhaps other storm surge examples and see if this change makes significant differences in results?


mjberger:
looks good to me.  — Marsha

mandli:
Will do.

mandli:
This looks fine to me.  We will just need to replace the tests.

rjleveque:
@mandli: OK, I will update the regression data for the surge test.

rjleveque:
After more testing, I think this PR is fine.  It makes some minor changes in results, particularly speeds (which are highly sensitive anyway), but I've rerun some big inundation models and nothing major changes.  And this is a bug that needs fixing.

PR #480, opened 2020-09-28
fetch_noaa_tide_data - handle missing data better

rjleveque:
Return None for water_level and/or prediction if it is not returned by fetch.

This works on my example, but fails in tests/test_util.py for reasons I don't understand.

Here's an example where I need this capability, for the gauge at Aberdeen, WA:

```
from clawpack.geoclaw import util
import datetime
begin_date = datetime.datetime(2020,9,13)
end_date = datetime.datetime(2020,9,18)
time, eta, etap = util.fetch_noaa_tide_data('9441187', begin_date, end_date)  
```
This throws an error, whereas with the new version it returns `time` and `etap` as arrays, which is all I need, with the message:
```                                                        
Fetching water_level data from NOAA for station 9441187
*** Fetching water_level failed, returning None
Fetching predictions data from NOAA for station 9441187
```


mandli:
Do you know specifically what the exception being raised is?  Might be something weird.

rjleveque:
I get the same error from nosetests on my laptop as travis is showing:

```
tests.test_util.TestFetchNoaaTideData.test_api_error ... Fetching water_level data from NOAA for station 1234567
*** Fetching water_level failed, returning None
Fetching predictions data from NOAA for station 1234567
*** Fetching prediction failed, returning None
FAIL
tests.test_util.TestFetchNoaaTideData.test_date_time_range_mismatch ... Fetching water_level data from NOAA for station 1234567
Fetching predictions data from NOAA for station 1234567
ok
tests.test_util.TestFetchNoaaTideData.test_missing_values ... Fetching water_level data from NOAA for station 1234567
Fetching predictions data from NOAA for station 1234567
ok
tests.test_util.TestFetchNoaaTideData.test_retrieve_and_cache ... Fetching water_level data from NOAA for station 1234567
Fetching predictions data from NOAA for station 1234567
Using cached water_level data for station 1234567
Using cached predictions data for station 1234567
ok
FAIL: tests.test_util.TestFetchNoaaTideData.test_api_error
  File "/home/travis/virtualenv/python3.6.10/lib/python3.6/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/virtualenv/python3.6.10/lib/python3.6/site-packages/nose/tools/nontrivial.py", line 67, in newfunc
    raise AssertionError(message)
AssertionError: test_api_error() did not raise ValueError
```

mandli:
@jkapilian, can you take a look at this?  Seems like there's some exception being raised that we are not handling correctly and it looks like things are failing ungracefully.

jkapilian:
@rjleveque it looks like tests/test_util.py is expecting it to throw a value error, but since you added a try-catch it doesn't get that and only gets the error message you now print. I imagine all you'd have to do is reconfigure the api_error test in test_util.py to expect that error message instead and it should pass the test (I had a similar issue when I altered a different part of fetch_noaa_tide_data)

mandli:
@jkapilian do you mean that the try-catch blocks are hiding the problem?

jkapilian:
@mandli It's not that the try-catch is hiding the problem but the test file seems to automatically give bad input to try to cause a value error and then is confused because the try-catch now handles it differently instead of throwing that error, show changing what the test expects should fix it.

mandli:
@jkapilian gotcha.  We probably should catch and then handle the problem better then.

rjleveque:
Maybe there's a better way to fix this test, but I switched it to just expecting it to return `None` instead of raising `ValueError` and now it seems fine.

@jkapilian, @mandil: ok to merge?

mandli:
I will leave it to @jkapilian.

PR #482, opened 2020-10-01
update regression data for changes to rpt2_geoclaw

rjleveque:
These changes go with the modification to the `rpt2_geoclaw.f90` transverse Riemann solver in https://github.com/clawpack/riemann/pull/159, so travis won't be happy with this PR until that is merged in and `clawpack/clawpack` is also updated.

Four of the tests needed updating in order to pass with the new transverse solve.

Some other tests passed even though the results changed because of the tolerance used, and I updated the regression data for those as well.  So there is new data in:

 - tests/bowl_slosh/regression_data
 - tests/chile2010_adjoint/regression_data
 - tests/dtopo1/regression_data
 - tests/multilayer/regression_data
 - tests/netcdf_topo/regression_data
 - tests/particles/regression_data
 - tests/storm_surge/regression_data

I also committed new versions of `claw_git_status.txt` in these directories so they are up to date with the data, and some directories were missing these.  I removed any `claw_git_diffs.txt` files from git since these get messy and shouldn't be needed.


rjleveque:
New regression data is ok, ready to merge.

PR #486, opened 2020-11-06
allow memsize to be set in setrun.py

rjleveque:
goes with clawpack/amrclaw#267.  Won't pass tests until that is merged and clawpack/clawpack updated.


rjleveque:
Tests now pass.

PR #487, opened 2020-11-10
Remove several setrun parameters, allowflag routine

rjleveque:
I implemented some changes several of us have discussed:

1. Unlike `clawpack/amrclaw/src/2d/flag2refine2.f90`, the routine in`src/2d/shallow/flag2refine2.f90` still called `allowflag` to check if cells are in "regions" where refinement is allowed or forced.  This is now done in `clawpack/amrclaw/src/2d/flagregions.f90` so it is not needed here.  Moreover the amrclaw version also checks new `flagregions` defined as  ruled rectangles, which the geoclaw version of `allowflag` does not, leading to errors.  Rather than fix `allowflag`, I removed it (from the directory and from `Makefile.geoclaw`).

2. It has always been possible to specify `minlevel` and `maxlevel` parameters associated with topo, dtopo, and qinit files, which implicitly created "regions" with x-y extent given by the topo/dtopo/qinit file.  This led to lots of confusion and has not proved particularly useful.  Those parameters are no longer supported.  If you want a region with the same shape as a topo file, you must now explicitly create a "region" or "flagregion".  For some reason these implicit regions were checked *both* in `flag2refine2` and in `allowflag`.  Now both sets of checks are gone.    

3. Another parameter that hasn't proved so useful and leads to confusion about where refinement is allowed was the `max_level_deep` and corresponding `deep_depth` parameters.  These have also been removed.  The new `flag2refine2` routine is greatly simplified as a result of all these changes.

4. I made corresponding changes in `src/2d/shallow/multilayer`, in `Makefile.multilayer`, `flag2refine2.f90`, and `qinit_module.f90`.

4. I modified the `setrun.py` files in a few examples to remove these parameters:
   - `examples/tsunami/chile2010`
   - `examples/tsunami/bowl-radial`
   - `examples/multi-layer/bowl-radial`
   - `examples/storm-surge/ike`

   Other examples should be modified once others approve of these changes.

Note that other examples still run fine, since specifying the deprecated parameters is still allowed and I don't think any of the examples actively used this implicit setting of regions.  If you continue to specify these parameters, you get warning messages such as:

```
/Users/rjl/git/clawpack/clawpack/geoclaw/data.py:185: UserWarning: 
  *** WARNING: topofile specs changed in v5.8.0 -- Flag level info now ignored
  warnings.warn(w, UserWarning)
/Users/rjl/git/clawpack/clawpack/geoclaw/data.py:130: UserWarning: 
  *** WARNING: deep_depth parameter ignored as of v5.8.0
  warnings.warn(w, UserWarning)
/Users/rjl/git/clawpack/clawpack/geoclaw/data.py:133: UserWarning: 
  *** WARNING: max_level_deep parameter ignored as of v5.8.0
  warnings.warn(w, UserWarning)
```

For some reason some of the regression tests no longer pass with these changes, which I don't understand and will investigate further before merging.

Note: Several pages in the docs need updating for these changes.


rjleveque:
@mandli: I notice in `flag2refine2.f90` the `wave_tolerance` and `speed` checks are only used in cells where `q(1,i,j) > dry_tolerance`, whereas the storm distance and wind speed checks apply to all cells, wet or dry, and so may flag more dry land for refinement than necessary.    Is this correct?

mandli:
There was a point where I thought a bit about this.  The problem is that the wind speed and radius checks are related to source terms so I think at the time I was thinking that it may have made sense to perhaps base these possibly on how close to the storm and wind speed magnitude these were in case something may be triggered.  In all honesty I am not entirely sure this makes sense though.

rjleveque:
@mandli:  `tests/particles` is the one that fails and I don't see why.  It passes on my laptop, also when I clone a brand new version of clawpack and merge in this PR.  Could you try running it locally when you get a chance?

Also, in `examples/storm-surge/isaac/setplot.py` I set `landfall = 0.` since this was not set and so gauge plots failed.  Is this the correct value?

rjleveque:
Travis tests pass after removing `tests/particles/__init__.py` so that one is no longer tested.

I am going to merge this and move forward, but at some point we should figure out what went wrong.  @mandli thinks it may be related to gfortran version issues, since newer versions of the compiler do not like several things in Clawpack.  We need to do some major work at some point...

PR #491, opened 2020-12-15
add a function util.bearing

rjleveque:
Computes the bearing from (x0,y0) to (x1,y1) on the sphere, sometimes
useful in conjunction with util.haversine, which computes great circle distance.

In particular this can be used to define a radially symmetric velocity field on the sphere as explained in the docstring.


rjleveque:
This just adds a new function and I'm going to merge so I can move ahead with other PRs.

But at some point it would be nice if someone tested and raised an issue if it's not done right!

PR #492, opened 2020-12-15
improve getmaxspeed.f90 for dry cells

rjleveque:
The subroutine `geoclaw/src/2d/shallow/getmaxspeed.f90` estimates the maximum wave speed on a patch, looping over all cells.  In dry cells it was doing a lot of arithmetic including sqrt's in order to do no update of the max value.  This routine is now simplified and should be faster on patches with lots of dry cells.

Also I noticed this on a problem where the depth `val(1,i,j)` was slightly negative and it was taking sqrt of a negative value, when it should have setting the wave speed to 0 in that cell.  I'm not sure how a negative value got in here since it should have been set to zero in `b4step`, but that's a different issue.  At any rate the new version won't have this potential problem.


mandli:
Makes sense to me.

mandli (approved):

PR #495, opened 2021-01-07
Fix force_dry_init indexes and tests

rjleveque:
This fixes a couple bugs in the `force_dry_init` array capability added in v5.7.0, see http://www.clawpack.org/force_dry.html

Array indexes were calculated improperly in `filval` due to not including ghost cells in the calculation, possibly leading to one or two cells near shore that should have been filled with water but are mistakenly forced to be dry initially.  In some cases this results in a significant wave being generated as water rushes into the dry cells from neighboring wet cells, so the shoreline appears to radiate nonphysical waves.  This showed up in some gauge results in narrow waterways and bays that had highly-oscillatory behavior before the tsunami should have arrived.

There was also some unintended logic with .or. rather than .and. in testing if `(t <= tend_force_dry)`.  After that time, this array should be ignored.  

There was also some logic to try to apply these arrays only in cells where new grid patches were generated, not in previously existing grids.  This wasn't working properly and probably is not needed, but should perhaps be rethought at some point.

### DIFF
--- a/src/2d/shallow/filpatch.f90
+++ b/src/2d/shallow/filpatch.f90
@@ -59,7 +59,6 @@ recursive subroutine filrecur(level,nvar,valbig,aux,naux,t,mx,my, &
 
     real(kind=8) :: eta1old, eta2old
     real(kind=8) :: veta_init_c
-    real(kind=8) :: x,y
     integer :: ii,jj
 
     ! Scratch arrays for interpolation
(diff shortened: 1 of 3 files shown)


src/2d/shallow/filpatch.f90
```diff
@@ -327,20 +326,17 @@ recursive subroutine filrecur(level,nvar,valbig,aux,naux,t,mx,my, &
                                                              + eta2 * slope(2,i_coarse,j_coarse)
                     h_fine = max(eta_fine - aux(1,i_fine + nrowst - 1, j_fine + ncolst - 1), 0.d0)
 
-                    if (variable_eta_init) &
+                    if (variable_eta_init) then
```
mandli, line 329: How did this even compile?!

mandli:
Given that these routines are also in the multilayer code was this functionality ever added there?

src/2d/shallow/filpatch.f90
rjleveque: It was an old-fashioned one-line fortran if test,
    if (condition) statement
that was split over two lines, but it confused me too when I saw it, so I decided to make it clearer.

mandli: Oh, good point.

rjleveque:
No, the force_dry_init functionality hasn't been added to the multilayer code.  I guess it should be added for the top layer at some point.

mandli:
I doubt it would be too hard to add to the top layer as long as it was only the top layer involved.  Obviously if that was not the case it would be significantly more complex.  Is there complexity here or a divergence of code that would preclude doing this easily/quickly (note I am not volunteering you here)?

rjleveque:
In principle it should carry over fairly easily, although doing a quick diff on the files, it does seem like they have diverged a bit independent of these new features so it might take a bit of work to resync them. 

Note that if you do want to add this, you might consider adding the `variable_eta_init` capability at the same time, which affects mostly the same set of routines, see http://www.clawpack.org/set_eta_init.html

mandli:
I just worry that if we do not move on this somewhat quickly that the codes will diverge a bit too much.  In any case this should be merged in the meantime.  I will raise an issue with the idea that the multilayer code should be pushed into feature parity with the single layer code ASAP and probably before any other new features get added to this part of the single layer code if that sounds reasonable?

rjleveque:
Sounds good to me.  I'd be happy to work on this with you sometime.  But for now it would be good to merge this and move forward with the v5.8.0 release, since there are also several other bug fixes that have recently been merged and should be released.

Issue #498, opened 2021-01-15
Make the Synthetic Storm Surge Example "Interesting"


mandli:
Currently the synthetic storm surge example works but does not have a lot going on.  It would be nice to make this example a bit more interesting by doing things such as
1. Adding gauges
2. Adding examples as to how to create a synthetic track and intensity for the storm
3. Create synthetic bathymetry
These are just suggestions of course.

As this example has been removed this issue now either needs to add this example back or should move what was here back into the apps repository.

rjleveque:
@mandli, I was remaking the galleries for v5.8.0 and discovered that this example now dies with "too many dt reductions" at `t = 5638.52`.  I see you recently changed it from order 1 to 2, and even with the new `rpt2_geoclaw.f` routine it dies for me (but still runs with order 1).  Did you get this to run with order 2?

mandli:
Odd, I thought it was working before with order 2 when I was running it before but I agree, it seems to not be working although I don't see why right now.  I will work on this a bit but if I don't see the issue right away we should roll back the order to 1 I suppose.

mandli:
Strangely when trying to debug this I increased the number of output and everything ran fine.

rjleveque:
Yes, increasing the number of output times seems to make it run, not sure why.  Also it takes hours to run, unlike most most things in the examples directories, and doesn't seem to illustrate anything in particular so I'm not sure what the goal is with this example.  For now I'm in favor of a quick fix so we can release 5.8.0, if possible.

mandli:
Hours to run?  For me it takes a couple of minutes, which admittedly is still longer than the other examples.  More importantly though is the stability of the problem either with more or less output steps.

As to what it was supposed to demonstrate, I was not aware that the example had the gauges removed.  It was supposed to show how to setup an example synthetic test case, which is something I constantly get asked about.  Right now this is pretty basic, which is why this issue was opened.  In principle this could be more a demo with a Jupyter notebook added with some more explanation.

rjleveque:
Am I looking at the right example?  In `geoclaw/examples/storm-surge/synthetic` the oldest commit is a6d4f1ef8d when it was renamed from `square-basin` in 2019, and there weren't any gauges in that version either.  It's a 3-level run from -12 hours to +3 days.

mandli:
That's the right one.  This is the timing I have with 3 outputs:
```
============================== Timing Data ==============================

Integration Time (stepgrid + BC + overhead)
[...]

PR #503, opened 2021-01-23
Remove examples/storm-surge/synthetic for now.

rjleveque:
Something strange is happening in this example, see clawpack/geoclaw#498.
It also needs some gauges and other enhancements to better illustrate
use of the storm surge code, and then perhaps put in the apps repository?

I discussed this yesterday with @mandli.


mandli:
There's actually an [example](https://github.com/mandli/surge-examples/tree/master/square_basin) that's similar already in the apps repository I think this one was originally based on.

Issue #504, opened 2021-01-25
Inaccurate documentation for iqinit = 4?


piyueh:
On GeoClaw's website (https://www.clawpack.org/setrun_geoclaw.html#setrun-qinit), it says

> iqinit : integer
> ...
> ... 4 = Perturbation to surface level

Maybe I misunderstand this description, but it seems the source code is simply taking the values as the water surface level `eta`, rather than the perturbation to `eta`. For example, in `qinit_module.f90`:

https://github.com/clawpack/geoclaw/blob/3431f6596985ecf39041ec798e67a4f86611288a/src/2d/shallow/qinit_module.f90#L183-L184

And also from here:

https://github.com/clawpack/geoclaw/blob/3431f6596985ecf39041ec798e67a4f86611288a/src/2d/shallow/qinit_module.f90#L168-L170

So I guess the documentation on GeoClaw's website is not accurate? (Actually, on another page [here](https://www.clawpack.org/topo.html#qinit-file), the description of `iqinit=4` in the last paragraph matches the source code.)

mandli:
You are correct that this is simply "perturbing" a state at rest but GeoClaw currently has no other way to initialize the sea-surface without additional code modifications so there really is no alternative anyway EXCEPT when an instantaneous earthquake is specified.  In that case I think it really does perturb the surface like you were thinking although I am not entirely certain, it was not really designed to be used simultaneously like that.

rjleveque:
@piyueh:  Yes, I think you're right that there are some inconsistencies in the documentation, and probably unintended behavior in the code in some cases.  Thanks for pointing this out.

This ability to read a qinit file dates from early versions of GeoClaw where we always assumed `sea_level == 0` and so any non-zero eta could also be viewed as the perturbation to 0.  I see that it needs some updating now that users are not only allowed to set a different constant `sea_level` but also a variable initial `eta` using the [`set_eta_init` function](https://www.clawpack.org/set_eta_init.html).

I think the lines you flagged in 
https://github.com/clawpack/geoclaw/blob/3431f6596985ecf39041ec798e67a4f86611288a/src/2d/shallow/qinit_module.f90#L168-L170
should probably be replaced by something like:
```
     else if (qinit_type == 4) then 
        eta = aux(1,i,j) + q(1,i,j)   # the surface elevation that was set in qinit.f90
        new_eta = eta + dq   # add the perturbation
[...]

PR #505, opened 2021-01-27
WIP: Fix how qinit_type==4 is handled: should be perturbation to eta from qinit

rjleveque:
Needed to update to handle case when original eta set in qinit is not just sea_level=0,
and also make more consistent with other types, so it perturbs eta only in wet cells, not in ones set to dry in qinit.

This implements my suggestion in #504.

Don't merge yet since there might be some discussion of this.

Also, @mandli should take a look at the multilayer case and see if we want to make a similar change there.


rjleveque:
Now I think the original code is correct, see #504.

PR #507, opened 2021-03-21
If no time offset is provided simply print none

mandli:
This was breaking when looking at the first time as this was just a float, which of course has no `isoformat()` response to give.


mandli:
Serves me right to rely on the github editor to actually not use tabs.  I will just open another PR that does not have a stupid editor.

PR #509, opened 2021-04-09
Fix reading of ensemble ATCF files

jkapilian:


mandli:
Can you mention real quick what this fixes?

jkapilian:
In my previous commit read_ATCF did not handle the time offsets correctly (this didn't affect regular best track files because they had all time information in the third column with an offset of 0 but ensemble files used the offset)

mandli:
There's a small conflict here that I think is not a big deal if you can fix it.

PR #510, opened 2021-05-06
Add geotiff to read for topotools

mandli:
Adds reading **only** of GeoTIFF files to the topotools `Topography` class.  This should allow someone to then convert a GeoTIFF topography file into another file supported directly by GeoClaw.

### DIFF
--- a/src/python/geoclaw/topotools.py
+++ b/src/python/geoclaw/topotools.py
@@ -761,6 +761,29 @@ class Topography(object):
                 if mask:
                     self._Z = numpy.ma.masked_values(self._Z, self.no_data_value, copy=False)
                     
+            elif abs(self.topo_type) == 5:
+                # GeoTIFF
+                try:
+                    import gdal
+                except ImportError as e:
+                    print("Reading GeoTIFF files requires GDAL.")
+                    raise e
+
+                data = gdal.Open(self.path)
+                z = data.GetRasterBand(1).ReadAsArray()
+                transform = data.GetGeoTransform()
+                x_origin = transform[0]
+                y_origin = transform[3]
+                dx = transform[1]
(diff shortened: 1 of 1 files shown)


mandli:
This does not yet have a test and should have one.  @rjleveque was the script example you sent me small enough that it could act as at least a functionality test (it does not fail)?

rjleveque:
@mandli: The script I sent downloads a 1/9" tile from the NCEI [CUDEM database](https://coast.noaa.gov/htdata/raster2/elevation/NCEI_ninth_Topobathy_2014_8483/)  as a test, but these are huge files (~200M).  

I think we should make a small tif to use as a test case, one that has the same format as the ones NCEI produces.  I'm not sure how to do this offhand -- so far I've only read them, not written one, but I can look into it.

mandli:
If we can add the functionality of also writing a GeoTIFF out that would be good too.

mandli:
Do you mean the output from the (design-a-grid site)[https://maps.ngdc.noaa.gov/viewers/grid-extract/index.html] or more generally?

Also note that you cannot get topography that crosses \pm 180 longitude anymore making it a bit more difficult to get trans-Pacific topography files.

rjleveque:
My comment from May 10 was incorrect and I think this is working properly, at least for .tif files from the NOAA [CUDEM database](https://coast.noaa.gov/htdata/raster2/elevation/NCEI_ninth_Topobathy_2014_8483/).

Other .tif files seem to use different conventions, and sometimes the `Z` array needs to be transposed (including those from the [grid-extract](https://maps.ngdc.noaa.gov/viewers/grid-extract/index.html) tool).

I'm still sorting this out, but will merge this for now and we can improve in the future perhaps.  Use with care.

Issue #512, opened 2021-06-09
Question about Holland 2010 implementation


tovogt:
Hey @bolliger32 and hey everyone,

I would like to understand some of the choices in your implementation of the Holland 2010 wind field model. It would be really great if you could answer the following questions:

1. There are some questions about the hardcoded parameters in the following lines: https://github.com/clawpack/geoclaw/blob/01c9a8e07cbb65b6bb21d8e5f79a09abecb1c4ae/src/2d/shallow/surge/model_storm_module.f90#L653-L658
1.1. Why is the peripheral radius `rn` set to 500 km (instead of the 300 km used in the Holland et al. 2010 paper)?
1.2. What is `rg`? Where does the formula for the peripheral exponent `xn` come from? It's not in the Holland et al. 2010 paper, right?
2. Why is there no Coriolis force term in the wind computation? https://github.com/clawpack/geoclaw/blob/01c9a8e07cbb65b6bb21d8e5f79a09abecb1c4ae/src/2d/shallow/surge/model_storm_module.f90#L675 For comparison, see how the Holland 1980 model incorporates the Coriolis force term `r * f / 2` https://github.com/clawpack/geoclaw/blob/01c9a8e07cbb65b6bb21d8e5f79a09abecb1c4ae/src/2d/shallow/surge/model_storm_module.f90#L592-L594
3. In Holland et al. 2010 (at the end of section 3.a.), it is stated that the "wind profile is not necessarily just the azimuthal flow;
if observations of the total flow are used, the profile will represent total winds (including the radial component)". So, why did you still choose to add the translational wind at the end?

Thanks in advance!
Thomas

mandli:
I actually think that @hudaquresh may have been the one who authored the Holland 2010 wind field.  I know that she did compare the output to make sure that it was producing good output but I am afraid I cannot directly answer your questions.  What I can probably say is the following:
1. Constants:
a. The outer radius of storms we usually use is read in if available so I am a bit surprised to find it hard coded here.  If it is not available we actually often set it to 500 km.  In the Holland 1980 wind model this is mostly used as a cutoff point though so it is not as critical as it is for this wind model.
b. Not sure why this is set to 10 m/s.
c. Not sure, looks like a scaling problem to me.  There are quite a few places that we have to convert units as well and this might be one of the corrections.
2. Looking at the Holland 2010 paper it does not look like this term was included.  I am not sure why it was not off-hand, maybe it should?
[...]

PR #513, opened 2021-06-14
Holland 2010: fix formula for peripheral exponent

tovogt:
Following the discussion in #512, this modifies the implementation of the Holland 2010 wind field model by fixing the formula for the peripheral exponent `xn`. Furthermore, some comments are added regarding the modeling choices.

One aspect from #512 that is not addressed here is the B-value computation: The original Holland et al. 2010 publication suggests to estimate the B-value directly from the central pressure and compute the maximum wind speed from there. Hence it suggests to disregard the max. wind speed measurement altogether. Implementing this in the Fortran code would require to add another storm variable: The time derivative of central pressure along the track. Are you also interested in this alternative B-value?

### DIFF
--- a/src/2d/shallow/surge/model_storm_module.f90
+++ b/src/2d/shallow/surge/model_storm_module.f90
@@ -604,6 +604,9 @@ contains
 
     ! ==========================================================================
     !  Use the 2010 Holland model to set the storm fields
+    !  
+    !  As in the original publication, this implementation computes
+    !  cyclostrophic winds, i.e., it does not include the Coriolis force.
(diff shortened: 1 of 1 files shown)


mandli:
It might be interesting to at least look at the differences in the B-values.  My guess again is that this was an oversight in the original implementation rather than a deliberate decision.

tovogt:
I decided to create a new PR for the updated B-value, see #520. Let's discuss this PR independent from the question of B-values.

mandli:
Thanks!

PR #518, opened 2021-07-06
Fix missing axes and non-long-lat plotting

mandli:
This PR incorporates two minor bug fixes to `Topography.plot()`:
 - Use the `axes` provided to the function if need be.  Before for the call to `pcolorcells` this was being ignored and only would work if the `axes` in question was the active one.
 - For non longitude-latitude plotting setting the aspect would not work as the code assumed longitude-latitude coordinates.  To fix this I added a new key-word argument that can be set to simply use `axes.set_aspect('equal')`, which more than likely should work at least for something like meters.


mandli:
Given that this seems to not be impacting anyone but a few I am going to merge this.

Issue #514, opened 2021-06-15
TCs in GeoClaw always rotate counterclockwise?


tovogt:
From running a toy example on the southern hemisphere I noticed that GeoClaw's wind field models seem to generate tropical cyclones with counterclockwise rotation regardless of hemisphere, even though rotation should be clockwise on the southern hemisphere. Can you confirm this?

mandli:
Oh, that is completely true.  Definitely a bug.

mandli:
Fixed by #519.

PR #519, opened 2021-07-12
Set orientation of TC according to hemisphere

tovogt:
This PR makes sure that the TC storm vortex rotates counter-clockwise on the northern and clockwise on the southern hemisphere (see #514).

### DIFF
--- a/src/2d/shallow/surge/model_storm_module.f90
+++ b/src/2d/shallow/surge/model_storm_module.f90
@@ -438,7 +438,8 @@ contains
     ! 5. Apply distance ramp to limit scope
     ! ==========================================================================
     pure subroutine post_process_wind_estimate(maux, mbc, mx, my, i, j, wind, aux, &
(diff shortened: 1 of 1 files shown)


mandli:
@tovogt I am wondering if we can do this more generally and simply have a sign value applied to the latitude?

tovogt:
Hm, I'm not sure what you mean. You can transfer the check `y >= 0` to a different place, but you will never get rid of it. For example, you can change the function `calculate_polar_coordinate` to set `theta = theta + 180` if `y < 0`. This will generate the same result in the end, but the function doesn't strictly compute polar coordinates any longer, which is really confusing if you need it for something else in the future.

mandli:
Yes, I guess that's what I was wondering instead of having it in each wind model instead.  Now that I look at what you did though more carefully I think you basically did this.

mandli:
Thanks BTW!

Issue #521, opened 2021-07-14
Are oscillations in gauge data unavoidable?


tovogt:
For testing purposes, I run GeoClaw with a synthetic tropical cyclone over open waters (constant bathymetry at -1000, no obstacles). It seems like even this simplistic example generates high oscillations in gauge output:
![image](https://user-images.githubusercontent.com/57705593/125652515-e21be643-3934-40cf-aaed-14429a769a03.png)
Is this kind of picture unavoidable from a mathematical (numerical) point of view? I also experience these kinds of oscillations in GeoClaw runs with real-world data and it doesn't seem to be very realistic. 

Here are the parameters (data files) I use to run this example:
[adjoint.data](https://github.com/clawpack/geoclaw/files/6817402/adjoint.data.log)
[amr.data](https://github.com/clawpack/geoclaw/files/6817403/amr.data.log)
[claw.data](https://github.com/clawpack/geoclaw/files/6817404/claw.data.log)
[dtopo.data](https://github.com/clawpack/geoclaw/files/6817405/dtopo.data.log)
[fgmax_grids.data](https://github.com/clawpack/geoclaw/files/6817406/fgmax_grids.data.log)
[fixed_grids.data](https://github.com/clawpack/geoclaw/files/6817407/fixed_grids.data.log)
[flagregions.data](https://github.com/clawpack/geoclaw/files/6817408/flagregions.data.log)
[friction.data](https://github.com/clawpack/geoclaw/files/6817409/friction.data.log)
[gauges.data](https://github.com/clawpack/geoclaw/files/6817410/gauges.data.log)
[geoclaw.data](https://github.com/clawpack/geoclaw/files/6817411/geoclaw.data.log)
[multilayer.data](https://github.com/clawpack/geoclaw/files/6817412/multilayer.data.log)
[qinit.data](https://github.com/clawpack/geoclaw/files/6817414/qinit.data.log)
[refinement.data](https://github.com/clawpack/geoclaw/files/6817415/refinement.data.log)
[regions.data](https://github.com/clawpack/geoclaw/files/6817416/regions.data.log)
[surge.data](https://github.com/clawpack/geoclaw/files/6817417/surge.data.log)
[topo.data](https://github.com/clawpack/geoclaw/files/6817418/topo.data.log)

mandli:
[...]

PR #523, opened 2021-08-25
coriolis: use absolute value for all wind field models

tovogt:
The very long discussion in #521 basically breaks down to this tiny detail: Even though it's never documented in the literature, the Holland 1980 wind field model requires to take the absolute value of the Coriolis parameter on the southern hemisphere. This also seems to apply to the CLE model (this is not mentioned in the CLE reference paper).

To make clear that this only affects the Holland 1980 and CLE models and no other wind field model, I removed the unused references to the Coriolis parameter in all other wind field models.


mandli:
This all checks out to me.

PR #524, opened 2021-09-19
Fix Floating Point Bug in Storm Track Time Comparison

mandli:
GeoClaw currently does not allow `t0` to be before the available track time.  There is therefore a floating point comparison that checks to see if `t0 < track_t0`.  Before this was implemented with
```fortran
t0 <= storm%track(1, 1) - TRACKING_TOLERANCE
```
which usually works except when `t0` and `storm%track(1, 1)` get large enough (not even really that large).  Instead this PR changes this comparison to
```fortran
t0 - storm%tracK(1, 1) < -TRACKING_TOLERANCE
```
which is safer from a floating point case.

### DIFF
--- a/src/2d/shallow/surge/model_storm_module.f90
+++ b/src/2d/shallow/surge/model_storm_module.f90
@@ -172,7 +172,10 @@ contains
             storm%central_pressure_change(storm%num_casts) = &
                 storm%central_pressure_change(storm%num_casts - 1)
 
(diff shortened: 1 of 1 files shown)


mandli:
Given the nature and simplicity of this fix I am going to merge this after testing this on a couple of additional storms.  If @bolliger32 or @tovogt have issues or anyone else who lives on the edge storm tracks and this does not work we may need to have a more elaborate way to check this or remove the requirement that we do not go before the first track time.  Suggestions for that are of course welcome.

PR #527, opened 2021-09-27
Allow for ATCF with missing data to be read

eht2126:
This change should allow for ATCF files with missing entries to be read in with np.nan filling the missing entries.


src/python/geoclaw/surge/storm.py
```diff
@@ -330,21 +329,32 @@ def read_atcf(self, path, verbose=False):
                 "TAU": lambda d: datetime.timedelta(hours=int(d)),
                 "LAT": lambda d: (-.1 if d[-1] == "S" else .1) * int(d.strip("NS ")),
                 "LON": lambda d: (-.1 if d[-1] == "W" else .1) * int(d.strip("WE ")),
+                #"BASIN":lambda d: (d.replace)
```
mandli, line 332: I think you may need to use `numpy.nan` here.

PR #531, opened 2022-01-10
OWI integration

cjeffr:
This PR is to add functionality to use data storm products such as Oceanweather (OWI) wind and pressure files for storm simulations. Example case will be sent separately as the current example is using proprietary data that can't be shared publicly.


mandli:
@cjeffr could you remove the extra files that only have a whitespace changes?  While I generally am fine with removing whitespace it looks like a bunch of files have been touched that are not related to the PR.

cjeffr:
Yeah that was weird, I have fixed it.

mandli:
@cjeffr is there a way we can a test or example for this added along side of this PR?

cjeffr:
The only example I have is one provided by the corps, if you have another data driven example I can modify it to make it work like OWI so that many people can use it.

mandli:
We should brain storm a way to have an example we can share and a test to make sure it works.  Something synthetic would work.

cjeffr:
One thing that immediately comes to mind is to just build a wind and pressure field using a hurdat or atcf example and the actual math like in geoclaw's original storm functionality but then save the data into the same format as the OWI netcdf file I created. Should be a reasonable approximation and can be tested against the original?

mandli:
Exactly what I was thinking.

cjeffr:
Ok I'll get started on that

examples/storm-surge/isaac_syn/setrun.py
```diff
@@ -0,0 +1,468 @@
...
+    data.display_landfall_time = True
+
+    # AMR parameters, m/s and m respectively
+    data.wind_refine = [20.0, 40.0, 60.0]
+    data.R_refine = [60.0e3, 40e3, 20e3]
+
+    # Storm parameters - Parameterized storm (Holland 1980)
+    data.storm_specification_type = 'OWI'  # (type 1)
+    data.storm_file = os.path.expandvars(os.path.join('/home/catherinej/storm_files',
+                                         'isaac_syn2.nc'))
```
YuchengZhang-hub, line 414: `data.storm_file = os.path.expandvars(os.path.join(os.getcwd(), 'isaac_synthetic.nc'))`

mandli:
@cjeffr Do you want to chat about this?  I just looked through to remind myself what changed but I have some easy suggestions that may be good to do.

mandli:
New changes will be pushed shortly, but I thought it good to document the performance differences between using a parameterized wind/pressure model and a data-derived one.  This is for the Isaac example and has not been fully verified yet.

|            | Wall   | CPU     |
| ---------- | ------ | ------- |
| OWI ASCII  | 36.537 | 154.782 |
| OWI NetCDF | 39.082 | 156.744 |
| Holland    | 12.715 | 43.252  |

This is inevitably due to the extra overhead of interpolation required, but does provide a base line that slow downs are not related to file I/O.

mandli:
@cjeffr can you pull in the PR to your branch?  I think this may be ready to merge in if so.

mandli:
I forgot to remove the `$CLAW/geoclaw/examples/storm-surge/isaac_ascii_data` and `$CLAW/geoclaw/examples/storm-surge/isaac_netcdf_data`.  I just pushed an update, or you can remove those directories.  Sorry about that.

mandli:
Also found a hard-coded path in the Makefile to `/opt/homebrew`.  Pushing that change as well (should be `$NETCDF4_DIR`).

mandli:
Probably not.  I had not realized that tests had even been added there.  At least the output should be removed.

PR #533, opened 2022-03-01
Add fgout grid capability and support for output_format binary32

rjleveque:
This (together with fgmax grids) replaces the old fixedgrid capability of GeoClaw, which has now been removed as an option.

This PR requires related PRs on  visclaw, and pyclaw in order for the visualizations to work in the example provided in `examples/tsunami/chile2010_fgmax-fgout`.

Some documentation will soon appear in the `dev` branch at https://www.clawpack.org/dev/fgout.html.

This is still being actively tested and there may be more updates.


rjleveque:
I still plan to add some more examples, but I think this is working well enough to merge after others review.

mandli:
@rjleveque I was attempting to test this and am running into problems with the `bowl_slosh` test.  It looks like it may have an update issue?

rjleveque:
This PR has been updated to also include support for specifying `output_format = 'binary32'` to write binary files with 4-byte/float32 values instead of the full 8-byte/float64 values that we compute with.  This results in files that are half the size but still have 8 significant figures, sufficient for all cases I have been working with, I believe.  When making lots of frames for an animation, this can be a significant reduction in storage.

In the process, I also added support for writing the standard output frames (all AMR grids) using `binary32` as well, which again cuts the file size in half and can be significant if writing many frames, or if doing many simulations. 

In `setrun.py` you can now specify `clawdata.output_format = 'binary32'`, or `'binary64'`.  For backward compatibility, if you specify `'binary'` you get `'binary64'`.  Of course you can also still specify `'ascii'` to get even larger files (but sometimes useful for debugging purposes).

An additional change is that the `fort.txxxx` files produced for each frame now includes a new final line with the `output_format`, so that when reading in the frames the proper format is used.   This avoids the need to set the `format` in `setplot.py` or to try to auto-detect it as in https://github.com/clawpack/visclaw/pull/222, and also fixes issues such as https://github.com/clawpack/visclaw/issues/275.

These changes to frame output are mixed in with the PR for new fgout capabilities because the same tools are used to read them in. For fgout grids the files produced are called `fgout.txxxx`, `fgout.qxxxx` and for binary output, `fgout.bxxxx`.  Some fixes to allow the proper `file_prefix` were also required in the other PRs listed below.

To test this PR, you also need:
- https://github.com/clawpack/pyclaw/pull/687
- https://github.com/clawpack/visclaw/pull/285
- https://github.com/clawpack/clawutil/pull/152

@mandli, the tests in `tests/bowl_slosh` work for me, but only when these other repos are properly checked out.

mandli:
The multilayer test is failing as are the examples.  I think it should be an easy fix as it looks like the Makefile just needs to be adjusted as the error during compilation is
```
/Users/mandli/Dropbox/src/clawpack/geoclaw/src/2d/shallow/multilayer/stepgrid.f:34:10:

   34 |       use fixedgrids_module
      |          1
Fatal Error: Cannot open module file 'fixedgrids_module.mod' for reading at (1): No such file or directory
```
that occurs in both the test and the examples.

rjleveque:
@mandli: the Makefile was already fixed to include `fgout_module` instead of `fixedgrids_module`, but I need to update `stepgrid.f` to use it properly.  Stay tuned.

rjleveque:
@mandli, please see if this looks ok now.

PR #536, opened 2022-04-23
Print gauges in a critical section, using a single unit number

rjleveque:
I changed the `print_gauges_and_reset_nextLoc` function to print all gauge data using the same output file unit number `OUTGAUGEUNIT` rather than creating a unit number based on this base plus the OMP thread number.  And I put this code in an OMP critical block so that only one thread can be doing this at a time.

The other approach allows writing in parallel, in principle, but recently I've been having seg faults when using lots of threads that seems related to this. It also seems potentially bad practice since we don't know what unit numbers will be used for writing gauges, and we use other unit number for other purposes, so there are potentially conflicts.

I don't think this will introduce any bottlenecks.

### DIFF
--- a/src/2d/shallow/advanc.f
+++ b/src/2d/shallow/advanc.f
@@ -241,7 +241,7 @@ c        # now to make linear interpolation easier, since grid
 c        # now has boundary conditions filled in.
 
 c     should change the way print_gauges does io - right now is critical section
-c     NOW changed, mjb 2/6/2015.
+c     NOW changed, mjb 2/6/2015.  RJL Changed back to critical block 4/22
 c     NOTE that gauge subr called before stepgrid, so never get
 c     the very last gauge time at end of run.
 
--- a/src/2d/shallow/gauges_module.f90
+++ b/src/2d/shallow/gauges_module.f90
@@ -675,14 +675,14 @@ contains
         integer, intent(in) :: gauge_num
 
         ! Locals
-        integer :: j, k, myunit
-        integer :: omp_get_thread_num, mythread
+        integer :: j, k
         character(len=32) :: out_format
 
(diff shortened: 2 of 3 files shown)


mandli:
There's a simple fix for this that does not require us to serialize this process.  Use a set number of unit numbers equal to the number of threads and reuse them.  We would therefore create a pool of units that would be available as required but would not go above the number of threads available.

rjleveque:
@mandli: Thanks for reviewing.  I think what you are suggesting is the way it was before I changed it: there was a unit number for each thread defined by 
```
   myunit = OUTGAUGEUNIT + mythread
```
But still I was running into segmentation faults for some runs with 20 threads that seemed to be caused by this, and some people use many more threads so we have to declare all unit numbers above 89 off limits for other purposes (since `OUTGAUGEUNIT=89`). 

Do you think this could be a bottleneck if it's serialized?

mandli:
...and it still causes you issues.  I guess I remember that you have mentioned this.  It definitely would not be great to serialize this if we have a large number of gauges but we also buffer the output to mitigate the problem as well.  Is this reproducible by chance?

In the end it is only serializing the output so maybe that's not so bad as long as we have buffers that avoid times when multiple gauges will be writing but I would expect that these are going to still overlap.  For instance, if a bunch of gauges are in the same area.  I would be better to understand why there is a seg-fault than anything.

rjleveque:
OK, we can hold off on merging this and I'll try to look into the seg fault issue some more.

mjberger:
i think it could be a bottleneck. People can use hundreds of gauges. Maybe we need a new approach.

 — Marsha

rjleveque:
This change was reverted, but I just opened a new issue #543 to remind us to revisit this at some point.

PR #538, opened 2022-07-28
fix the chile2010_fgmax-fgout example

rjleveque:
so its easier to make gallery files.  
`make fgout_plots` now makes the animations and moves them to `_plots_fgout`.


examples/tsunami/chile2010_fgmax-fgout/Makefile
```diff
@@ -63,10 +63,13 @@ topo:
 
 fgout_plots: 
 	$(MAKE) plots SETPLOT_FILE=setplot_fgout.py PLOTDIR=_plots_fgout
+	python make_fgout_animation.py
```
mandli, line 66: Do you want to use `$(CLAW_PYTHON)`?

rjleveque: Yes, thanks!

PR #545, opened 2023-02-04
savefig transparent param lower case

rjleveque:
This didn't used to throw an error but now it does.

### DIFF
--- a/src/python/geoclaw/kmltools.py
+++ b/src/python/geoclaw/kmltools.py
@@ -1385,7 +1385,7 @@ def kml_build_colorbar(cb_filename, cmap, cmin=None, cmax=None,
         
         
     # This is called from plotpages, in <plotdir>.
-    plt.savefig(cb_filename,Transparent=True)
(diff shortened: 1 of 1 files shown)


mandli:
Should this be somewhere else as well in PlotClaw?

rjleveque:
It's also in `visclaw.geoplot.kmltools.kml_build_colorbar`, but that routine has the docstring:
```
This version is deprecated, please use
        clawpack.geoclaw.kmltools.kml_build_colorbar
```
But we should probably update it in case it's used in legacy code.  Or just remove it at this point.

I've forgotten why we put `geoplot.py` in `visclaw` rather than `geoclaw`.

mandli:
So this is only for the transparency to overlay it onto Google Earth?  Probably not an issue then.

> I've forgotten why we put geoplot.py in visclaw rather than geoclaw.

I have been wondering that as well but it is never completely clear where to put this stuff sometimes.

rjleveque:
Yes, it's only used on Google Earth.

PR #555, opened 2023-08-07
fix typo in an error message

hydrocoast:
A very trivial correction but typo in fgout_tools.py: `binary63` to `binary64`


rjleveque:
Thanks for catching this!

Issue #553, opened 2023-07-28
FGmaxData.read: attribute "fgmax_grids" is not read in


tovogt:
When using the function `FGmaxData.read` on a `"fgmax_grids.data"` file that has been previously written using the function `FGmaxData.write`, the `fgmax_grids` attribute will remain empty. This is because the function `clawpack.clawutil.data.ClawData.read` only considers lines with a `=:` specification, while `FGmaxData.write` uses its own format to store the `fgmax_grids` attribute.

I would really love to have an implementation of `FGmaxData.read` that restores the original `fgmax_grids` attribute.

PR #556, opened 2023-08-15
Add FGmaxData read function

mandli:
Addresses #553 by adding a `read` function to the `FGmaxData` object.  Also modifies `fgmax_tools` to avoid a circular import problem.


tovogt:
Thanks, works great!

rjleveque:
Works for me too, thanks!

PR #559, opened 2023-08-29
topotools.Topography.__init__ now calls read() if path given.

rjleveque:
The arguments were also re-ordered so `topo_type` follows `path`, since this is the most common use case (`topo_func` is now third). So you can do:

    topo = topotools.Topography(filename, 3)

to create a `Topography` object and read data from filename with `topo_type=3`.

Before, you had to do

    topo = topotools.Topography()
    topo.read(filename, 3)

Not a major change but after years of doing this, I'm tired of it.

This is also now consistent with `dtopotools`, where you can do something similar.

### DIFF
--- a/src/python/geoclaw/topotools.py
+++ b/src/python/geoclaw/topotools.py
@@ -417,7 +417,7 @@ class Topography(object):
         return self._delta
 
 
-    def __init__(self, path=None, topo_func=None, topo_type=None, 
+    def __init__(self, path=None, topo_type=None, topo_func=None, 
                        unstructured=False):
         r"""Topography initialization routine.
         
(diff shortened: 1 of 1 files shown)


rjleveque:
Thanks @mandli. You can of course still instantiate with

    topo = topotools.Topography()

to create an object without reading a data file, and then you could set

    topo.path = filename
    topo.topo_type = 3

without yet reading it.

But perhaps you want an option where you can specify the topofile name and type when instantiating, without yet reading it, e.g.

    topo = topotools.Topography(filename, topo_type, read_data=False)

That would be easy to add, but I'm wondering what the use case is for that?

rjleveque:
I will merge this as is and we could add more options later.

PR #560, opened 2023-08-30
dtopotools: change 'is' to '==' to avoid a warning

rjleveque:
Very minor change.

### DIFF
--- a/src/python/geoclaw/dtopotools.py
+++ b/src/python/geoclaw/dtopotools.py
@@ -1131,7 +1131,7 @@ class Fault(object):
         is set to static.
         """
 
-        if self.rupture_type is 'static':
+        if self.rupture_type == 'static':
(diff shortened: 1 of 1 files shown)


mandli:
I always make this mistake with strings!

mandli:
In case anyone is curious about the difference someone explained it nicely as 

> You use `==` when comparing values and `is` when comparing identities.

and `==` implies `is` while the opposite is not true (except for `NaN`s for some reason).

rjleveque:
I think `==` compares the values whereas `is` tests if they are the same object, see e.g.
https://towardsdatascience.com/whats-the-difference-between-is-and-in-python-dc26406c85ad

One thing I read said to use `is` for testing if something `is None`, but otherwise `==` is usually what you want.

PR #562, opened 2023-09-01
WIP: Rename regression_test.py to use pytest

rjleveque:
See also https://github.com/clawpack/classic/pull/90

Gives 4 errors currently, 3 of which are related to precision of results.

```
=========================== short test summary info ============================
FAILED test_storm.py::test_storm_IO - TypeError: 'DataArray' object cannot be interpreted as an integer
FAILED bowl_slosh/regression_test.py::BowlSloshTest::runTest - AssertionError: 
FAILED dtopo1/regression_test.py::DTopoTests::runTest - AssertionError: 
FAILED particles/regression_test.py::ParticlesTest::runTest - AssertionError: 
============ 4 failed, 26 passed, 2 skipped, 16 warnings in 41.24s =============
```

@mandli: maybe you can figure out what's wrong with `test_storm.py` and I'll look into the other failures. It seems like the results agree to about single precision but not to the specified tolerances.


rjleveque:
The disagreement with regression data is probably due to some changes in how regions are applied and hence what AMR grid patches are used. There are similarly small changes for a couple of amrclaw tests.

I updated the `regression_data` directories so that tests should all pass, except for `test_storm.py`

mandli:
Superseded by #642.

PR #567, opened 2023-09-17
use CLAW_PYTHON instead of `python` in Makefile

hydrocoast:
Similar to #538 and as discussed in https://github.com/clawpack/clawutil/pull/156  
Just replacing and no breaking change.


rjleveque:
Good idea, thanks @hydrocoast!

PR #570, opened 2023-10-10
Add sphere_source parameter

rjleveque:
So that sphere source terms can be selected in setrun.py by setting
`rundata.geo_data.sphere_source`

Added to data.py, geoclaw_module.f90 and used in src2.f90.

For now set default to 0 for backward compatibility, but default
should eventually be 1 so that source term in mass is included.


mandli (commented):
Are you thinking of allowing multiple methods to be chosen?  Otherwise it seems like a `True` or `False` would be more clear.

rjleveque:
@mandli, Yes for now at least it can be 0, 1, or 2 where 2 means add source terms in momentum as well as mass.  But those are nonlinear in velocities and typically negligible for practical problems, I believe.  I'm working on some examples...

PR #574, opened 2023-11-01
Add libraries for 1d version on single grid, and examples

rjleveque:
This adds new functionality and shouldn't affect existing code.

Added:
- `src/1d_classic/shallow`: Source code for shallow water equations
- `src/1d_classic/bouss`: Source code for Boussinesq version
- `examples/1d_classic/`: Directory with 6 examples
- `src/python/geoclaw`: Some new modules and updated `dtopotools.py`

Requires updates in other repos:
- [clawutil#158](https://github.com/clawpack/clawutil/pull/158) for updated `data.py` used in `setrun.py` functions
- [visclaw#294](https://github.com/clawpack/visclaw/pull/294) for updated `geoplot.py` used in `setplot.py` functions.

Notes:
- AMR is not supported, only a single (nonuniform) grid.  
- The function `geoclaw.nonuniform_grid_tools.make_celledges_cfl` can be used to create a grid that has Courant number close to 1 in deep water, transitioning to a uniform fine grid onshore.
- The Boussinesq solvers include Madsen-Sorensen (with B=1/15 by default) and the version of SGN equations suggested by Bonneton et al. with alpha = 1.153 by default.
- More info about the examples and the plots they produce can be viewed at https://depts.washington.edu/clawpack/sampledocs/geoclaw_1d_classic_examples/
- Some documentation with more details can be found at https://www.clawpack.org/dev/geoclaw1d.html
- Based on development in a separate `clawpack/geoclaw_1d` repository up through https://github.com/clawpack/geoclaw_1d/pull/3.  See that repo for more development history.


mandli:
Very minor organization comment.  Why not just label this as 1d like the rest of the code?  If there is another 1d code maybe contain it below that?

rjleveque:
@mandli: There is already a `src/1d` directory that contains a `multilevel` subdirectory, although I'm not sure if that's up to date and I don't think any examples use it.  Also, since this new code doesn't work with AMR I thought it might be better to separate it out as `1d_classic` so that later perhaps we could put code in `1d` that extends what's in `amrclaw/src/1d` to work for geoclaw.

rjleveque:
Now that v5.9.2 is out I will merge this in, to be included in v5.10.0.  It may need some more updates / additional examples so please continue to test and let me know about any issues.

PR #579, opened 2023-11-14
Topo module variables removed from cellgridintegrate call sequence

rjleveque:
WIP: Still need to test some more before merging.

Remove `topo_module` variables from calling sequence of `cellgridintegrate`
    
They don't need to be in the calling sequence since setaux already assumes they are in topo_module. 

Some integers were implicitly declared as (kind=4) whereas recent changes to topo_module.f90 in #540 declared them as (kind=8) to deal with larger topo files.  As a result the code dies when reading some large topo files when i0topo overflows 4 bytes and was set to 0 in `gridcellintegrate`.

@bolliger32: You might want to take a look and see if it still works for your big files.

### DIFF
--- a/src/2d/shallow/cellgridintegrate2.f
+++ b/src/2d/shallow/cellgridintegrate2.f
@@ -1,10 +1,8 @@
 c====================================================================
-      subroutine cellgridintegrate(topoint,xim,xcell,xip,yjm,ycell,yjp)
+      subroutine cellgridintegrate(topoint,xim,xip,yjm,yjp)
 c=====================================================================
 
-c *** Removed topo_module variables from calling sequence 
-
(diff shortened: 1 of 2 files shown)


bolliger32:
@rjleveque sorry I didn't get to this before the merge but will test out on the big files. Sorry for not catching this in #540 . Changes in cellgridintegrate make sense to me. The diffs seem to be messed up for setup so didn't compare those but regardless, I'll test it out!

rjleveque:
Yes, sorry about diffs in setaux, I discovered that Windows CRLFs had been put at the ends of lines sometime years ago, and it differed from the version elsewhere, so I used dos2unix to clean it up.  But it was only the calls to cellgridintegrate that were changed.

bolliger32:
Ahh makes sense. Good thing to fix

PR #581, opened 2023-11-29
Add nonuniform_grid_tools to meson.build

carlosmunozmoncayo:
@ketch found a fix to the module _nonuniform_grid_tools.py_ not being imported with
`from clawpack.geoclaw import nonuniform_grid_tools`.

### DIFF
--- a/src/python/geoclaw/meson.build
+++ b/src/python/geoclaw/meson.build
@@ -13,6 +13,7 @@ python_sources = {
     'kmltools.py',
     'marching_front.py',
     'most2geoclaw.py',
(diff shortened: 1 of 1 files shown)


ketch:
@rjleveque @mandli With our new build system (meson), any python file that is to be importable must explicitly be listed in the `meson.build file`.  So typically when any new Python file is added to the code, there should be a corresponding line added to `meson.build`.

mandli:
Was this being called from somewhere that broke?  Obviously it would be best to implement this fix but I do not actually remember anywhere that it is used directly.

PR #599, opened 2024-04-14
change nodata_value to -99999 in topotools.py

rjleveque:
Previous value `-9999` did not have enough digits since new etopo 2022 data has this as an actual topography value in the Mariana Trench (when written with `Z_format='%.0f'`).


mandli:
Heh, was wondering when we might need to do this.

PR #601, opened 2024-04-19
Add Plane-Wave Storm

mandli:
This adds a new model storm type that is simply a pressure core of a Holland '80 storm but a plane-wave, i.e. it extends in the y-direction.  This is useful for testing purposes and so was decided to leave it in for such things.  Future parameters could include an angle for the symmetry or at least direction so that symmetry can be placed in the x-direction as well.


mandli:
I am going to drop this as I think it is a bit too niche and can mostly be accomplished via a synthetic storm.  Miscellaneous improvements that were part of this PR that are relevant have been added to #616.

PR #602, opened 2024-04-27
Improve chile2010_fgmax-fgout example script for fgout animation

rjleveque:
This script has been simplified since I realized that `update_artists` are defined globally and do not need to be passed into `update`, unpacked, and repacked.  Moreover if `blit==False` this tuple is not needed at all.  This does not seem to slow it down at all.  Getting rid of `update_artists` makes it easier to adapt to other animations where many things need to be updated in each frame.

I also cleaned up several other things in the script and added a better doc string.


rjleveque:
I just added a new example script `make_fgout_animation_with_transect.py` to show how to also plot a transect of the solution and topography from fgout data.

mandli:
I like that this example actually uses the "other figure" functionality.  Not sure that we had that before.

PR #604, opened 2024-05-01
handle 0 radius

bolliger32:
This addresses an issue in the edge case where `r`, the distance of the grid cell centroid to the storm eye, is 0. Previously this would result in either floating point errors or nan's in the `q` array, depending on the compile flags.

[Google group discussion](https://groups.google.com/g/claw-users/c/BPAfEZZTrMI/m/8oE1idBdBgAJ)

### DIFF
--- a/src/2d/shallow/surge/model_storm_module.f90
+++ b/src/2d/shallow/surge/model_storm_module.f90
@@ -643,7 +643,7 @@ contains
 
                 ! Set speed of wind at this point, handling case of grid cell centroid
                 ! at eye of storm
-                if (r == 0) then
+                if ((mwr / r)**B > 100) then
                     wind = 0
                 else
                     wind = sqrt((mwr / r)**B &
@@ -715,7 +715,7 @@ contains
 
                 ! Set speed of wind at this point, handling case of grid cell centroid
                 ! at eye of storm
-                if (r == 0) then
+                if ((mwr / r)**B > 100) then
                     wind = 0
                 else
                     wind = sqrt((mwr / r)**B &
@@ -803,7 +803,7 @@ contains
 
                 ! Set speed of wind at this point, handling case of grid cell centroid
                 ! at eye of storm
-                if (r == 0) then
+                if ((mwr / r)**B > 100) then
                     wind = 0
                 else
                     ! # HOLLAND 2010 WIND SPEED CALCULATION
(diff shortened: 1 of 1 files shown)


rjleveque:
When `r` is very close to zero, `exp(-(mwr / r)**B)`  may underflow.  Is that a concern? If so, the tolerance should be chosen appropriately.

bolliger32:
@mandli I updated that comparison to what you showed. My hunch on the pressure is that maybe it makes sense to cross that bridge when we need to, e.g if we have a case where we want to use a different pressure field function.

@rjleveque do you think 1d-16 is too restrictive? Since these are all 8-byte floating points, I think it would take a pretty big value of B and/or mwr for `-(mwr / r)**B` to give an underflow error if `r=1d-16`. The caveat being I don't totally know how underflows occur...

rjleveque:
Maybe I'm missing something, but even for r = 1e-2, exp(-1/r) < 1e-43 and when r = 1e-3 it underflows.  For r = 1e-16 the exponent of exp(-1/r) has 16 digits.  With 8-byte floats the exponent only holds 8 bits, so 3 digits at most.

It seems like you might want to do something like:
```
        if (-(mwr / r)**B < -100) then
            pres = Pc
        else
            pres = Pc + dp * exp(-(mwr / r)**B)
        endif
```

rjleveque:
It's generally not a problem but depending on what compiler flags someone uses, it might throw an error.

bolliger32:
@rjleveque ok makes sense! I've incorporated your suggestion.

mandli:
You may want to do something similar for the wind evaluations.  While $r \rightarrow 0$ goes to 0 it would be safer to force the issue for small $r$.

bolliger32:
@mandli good point. I added the same if/then statement for the wind evaluations.

mandli:
This looks good to me.  I am not sure what threshold makes sense but I would be surprised if this was not sufficient.  @rjleveque do you have any thoughts before merging?

rjleveque:
I don't know what threshold is best. Since `exp(-100) = 3.7e-44`, you could make 100 something smaller, but I'm not sure if it matters unless elsewhere these small numbers get multiplied by other equally small numbers, which then causes underflow.

mandli:
I do not think that will be a problem so let's just go with this.

PR #607, opened 2024-05-08
Update Makefile.bouss to point to riemann/src for Riemann solvers

rjleveque:
Requires updated solvers to handle the fact that `q` and waves, fluctuations now have 5 components rather than 3.  This is addressed in https://github.com/clawpack/riemann/pull/174, so this PR will not work without that one.


rjleveque:
Note that this PR means that `$CLAW/riemann/src/rpt2_geoclaw.f` will be used in place of the version that was in `src/2d/bouss/rpt2_geoclaw_sym.f`.  But that was an old version, branched from before https://github.com/clawpack/riemann/pull/159 was merged, and the riemann repository version should be better.  

Making this switch makes virtually no differences to the results computed for several Boussinesq examples I've tested.

PR #615, opened 2024-05-30
Gauge filenames

rjleveque:
This addresses https://github.com/clawpack/geoclaw/issues/541, creating filenames in the same format as previously if the gauge number has 5 or fewer digits, e.g. `gauge00012.txt`, but allows up to 15 digits, e.g. `gauge123456789.txt`.

If this seems ok, I'll do the same thing in amrclaw.

Note that gauges are still read in fine for plotting regardless of the number of digits, with no changes needed in `pyclaw.gauges.GaugeSolution`.

Note that this is implemented via these two lines (where `numstr` is a character variable):

```
write (numstr,'(I0.5)') num
gauges(i)%file_name = 'gauge'//trim(numstr)//'.txt'
```

This is a much cleaner way to do it than the old way of starting with `gaugexxxxx.txt` and changing one character at a time!

We should eventually make this change also in creating filenames like `fort.t0001`, in `valout.f90`, both for cleaner code and greater flexibility if the user needs more digits than we originally planned for.

### DIFF
--- a/src/2d/shallow/gauges_module.f90
+++ b/src/2d/shallow/gauges_module.f90
@@ -207,18 +207,12 @@ contains
             end do
 
             ! Create gauge output files
-            ! with format gauge00012.txt or gauge1234567.txt
             do i = 1, num_gauges
                 num = gauges(i)%gauge_num
-                ! convert num to string numstr:
-                if (num > 100000) then
-                    ! > 5 digits,  no padding
-                    write (numstr,'(I0)') num
-                else
-                    ! <= 5 digits, add zero padding
-                    write (numstr,'(I5.5)') num
-                endif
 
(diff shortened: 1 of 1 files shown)


mandli:
Works for the examples I tested.

rjleveque:
Thanks @mandli.

In further testing I realized that the maximum gauge number is still limited by the fact that this is read in and stored as an 8-byte integer, so not larger than 2147483647.  But at least it now allows any 9-digit or shorter gaugeno.

mandli:
I noticed that but I figured we can work out the warning/error message when added to AMRClaw where the Python data stuff is handled.

rjleveque:
Good idea, I'll modify the amrclaw PR I just did...

PR #617, opened 2024-06-06
improvements to fgout_module for t arrays and D-Claw

rjleveque:
Allow specifying `fgout.output_times` as a general list or array of times rather than equally spaced, if you set `fgout.output_style = 2`.  The default is `fgout.output_style = 1` that takes the previous input. Changes to `fgout_tools.py` allow specifying these values in `setrun.py`.

Then we also made this more general so that it could handle D-Claw output as well as GeoClaw, with a new attribute `fgout.dclaw` that is `False` by default, but setting this to `True` outputs 7 components of `q` instead of only 3 (plus `eta` in both cases).

I plan to also modify this so that you can specify fewer components rather than all of them. 

Also cleaned up a bunch of trailing white space.

### DIFF
--- a/src/2d/shallow/fgout_module.f90
+++ b/src/2d/shallow/fgout_module.f90
@@ -10,7 +10,7 @@ module fgout_module
         real(kind=8), pointer :: late(:,:,:)
 
         ! Geometry
-        integer :: num_vars(2),mx,my,point_style,fgno,output_format,output_style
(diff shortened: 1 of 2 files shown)


rjleveque:
I plan to merge this for now, and work on additional updates later - see https://github.com/clawpack/geoclaw/issues/618

@kbarnhart, let me know if you run into problems using this.

kbarnhart:
@rjleveque Thanks! Will let you know if I run into issues.

PR #621, opened 2024-06-19
fix fgout_tools.FGoutGrid.read_fgout_grids_data for time array

rjleveque:
Now that `output_style==2` is supported for fgout grids (an array of times, see https://github.com/clawpack/geoclaw/pull/617), the function fgout_tools.FGoutGrid.read_fgout_grids_data needs to be fixed to properly read in the new format of `fgout_grids.data`.  And also note that `nout` now comes before `tstart` and `tend` for `output_style==1`.

I was thinking we might might want to rewrite this code so that it makes a `.data` file with `:=` in the lines, similar to `claw.data` and others, so that `clawutil.data.ClawData.read` can be used to read it rather than depending on a particular line ordering.  But the problem is that there might be more than one fgout grid specified in the file, so I don't think this will work. (Similarly for fgmax).

### DIFF
--- a/src/python/geoclaw/fgout_tools.py
+++ b/src/python/geoclaw/fgout_tools.py
@@ -326,11 +326,32 @@ class FGoutGrid(object):
             raise ValueError('fgout grid fgno = %i not found in %s' \
                              % (fgno, data_file))
 
-        self.tstart = float(fgout_input[0].split()[0])
-        self.tend = float(fgout_input[1].split()[0])
-        self.nout = int(fgout_input[2].split()[0])
-        self.point_style = point_style = int(fgout_input[3].split()[0])
-        output_format = int(fgout_input[4].split()[0])
+        lineno = 0 # next input line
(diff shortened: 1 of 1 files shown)


mandli:
Is it possible that you would know a priori how many grids and then read each?  That's how some of the other data files work.

rjleveque:
The first thing in `fgout_grids.data` is the number of fgout grids, and the current implementation first reads that and then loops over the grids to read the other data.  The problem is that it wouldn't work to replace the line 
1                           # fgno
by
1                           := fgno
for example, and expect to read in the grid info properly with `clawutil.data.ClawData.read` if the file also has a later line 
2                           := fgno
for a second fgout grid.

Short of having a separate `.data` file for each fgout grid, best avoided, I don't think it would be easy, or worthwhile to spend time on this.

I added the comment since I know I've thought about this before and others may have wondered about this too.

mandli:
I guess I was thinking of writing a read routine rather than making it automatic, which would be difficult to finagle.

PR #624, opened 2024-07-14
Allow specifying which components of q to output on each fgout grid

rjleveque:
Modifications to fgout_module.f90 and fgout_tools.py, primarily. Can now set `fgout.q_out_vars` to a list of which components of q to output for each fgout frame (Fortran indexing). The default is `fgout.q_out_vars = [1,2,3,4]` for all components of q[1:3] and also eta (4), consistent with the previous behavior on standard GeoClaw.

The user could also/instead ask to output the topo B as component 5.  (If two out of three of h, eta, B are output then the other can be computed from these.)

Note that this list is written out to `fgout_grids.data` rather than a boolean list of True/False values for each possible component as is done in other places in GeoClaw. This turned out to be both easier and clearer.

I got rid of the `fgout.dclaw` parameter that was introduced in #617 and all the complication of hard-wiring different sets of variables for geoclaw vs dclaw since this should now allow much more control to use either as well as reducing the size of fgout files when only a few components are desired.

For the Boussinesq version of GeoClaw, there are 6 variables in q and so to print out only h,hu,hv,eta you would specify `fgout.q_out_vars = [1,2,3,7]`, or just `[1,7]` if you do not want to write the momenta and only want h,eta.

Needs some more testing still and I'd be happy for feedback.


rjleveque:
This has been updated to handle Bouss and D-Claw.
 
When instantiating `fgout_tools.FGoutGrid` object, new qmap parameter can be set
to 'geoclaw', 'geoclaw-bouss', 'dclaw', or a custom dictionary.
(Eventually add 'dclaw-bouss' option also). 
This dictionary has the form `{'h':1, 'hu':2, etc}` with mapping from fgout
variable names to Fortran indices of q in the code creating the fgout output.

rjleveque:
I fixed a few issues and I think this is ready to merge.  I also updated the documentation, see
  https://www.clawpack.org/dev/fgout.html

PR #625, opened 2024-07-18
Update topo_module.f90

mandli:
The `character` variables that were storing variable names and IDs for NetCDF topography reading were only length 10.  This just makes them 64 characters long.  Probably too long but not really a big deal.


rjleveque:
Looks fine.

PR #630, opened 2024-10-21
Properly set fgout q_out_vars to default

rjleveque:
Should be `[h,hu,hv,eta]` as in old versions, but now `qmap['eta']` varies depending on equations being solved (e.g. `geoclaw-bouss` or `dclaw`).


mandli:
I am not sure I am understanding why `qmap` would not be defined and would require a try-except block.  If anything maybe we are expecting an integer in the dictionary?

rjleveque:
`qmap` should always be defined but the user might pass in a custom `qmap` that does not contain `eta`.  Probably not, but I was being extra cautious.

The main change is that `qmap['eta']` should be used rather than hardwired to `4`.

mandli:
Maybe just checking the dictionary for the key `eta` and throw an exception then?

rjleveque:
> Maybe just checking the dictionary for the key `eta` and throw an exception then?

Good idea to check for key, I changed it.

PR #634, opened 2024-12-15
FIx filename inconsistency in examples/tsunami/chile2010_fgmax-fgout

hydrocoast:
I dropped the “_with” in the file name to make it executable with `make plots_fgout`.
Anyway, the transection view is a really great tool!


mandli:
Thanks!

PR #636, opened 2025-02-10
modify so gauge values always written at start time regardless of min_time_increment

rjleveque:
Previously if `min_time_increment > 0` was set for a gauge it wouldn't capture the gauge value until this long after the starting time value `t1` specified for the gauge.


rjleveque:
By the way, I also tried to make it print the gauges at the final time in the calculation, since currently the gauge output is done _before_ the call to stepgrid on each grid.  But it's done that way so that ghost cell values have been filled in already, which in general are needed if interpolation is used to get the gauge value.  (In practice we try to insure that gauges are at the centers of grid cells on the finest level but that's not always practical.). Anyway, I don't see any easy to fix this.

PR #638, opened 2025-03-01
fix bugs in willoughby wind model

sustielz:
Changed the maximum wind radius (mwr) in the coefficients a and N to match units in the original Willoughby paper, and fix a sign error in the parameter W

### DIFF
--- a/src/2d/shallow/surge/storm_module.f90
+++ b/src/2d/shallow/surge/storm_module.f90
@@ -113,6 +113,7 @@ contains
         use model_storm_module, only: set_rankine_fields
         use model_storm_module, only: set_modified_rankine_fields
         use model_storm_module, only: set_deMaria_fields
+        use model_storm_module, only: set_willoughby_fields
 
         ! use data_storm_module, only: set_data_storm => set_storm
         use data_storm_module, only: set_HWRF_fields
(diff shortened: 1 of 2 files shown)


mandli:
This does not completely enable the Willoughby model but gets the changes we need in.  Can you 
- [ ] Add the necessary python to allow willoughby to be used.  Line 551 in `$CLAW/geoclaw/src/python/geoclaw/data.py`.
- [ ] Add the DOI URL (https://doi.org/10.1175/MWR3106.1) instead of the current URL

I can also merge this in as it has the essentials and then do another PR that enables the wind model.

mandli:
Ideally we would have a full on test.  If you would feel more comfortable now enabling it in the mean time I am fine with that as well.  Would be good to change the URL though.

src/2d/shallow/surge/storm_module.f90
```diff
@@ -228,6 +229,8 @@ subroutine set_storm(data_file)
                         set_model_fields => set_modified_rankine_fields
                     case(7) ! deMaria model
                         set_model_fields => set_deMaria_fields
+                    case(9) ! Willoughby model
```
mandli, line 232: Need to change line 214 above to be 9.

mandli:
Thanks @sustielz!  All looks good.

PR #639, opened 2025-03-13
Draft: topoarea() fix?

kbarnhart:
@rjleveque , @mandli - I'm working on a synthetic problem and came across a *potential* bug in the topoarea calculations. 

In my application I have created a topo grid that perfectly aligns with my simulation grid. When I did this, I ran into topoarea errors indicating that my topography was not big enough for the domain. I think this occurred because the topoarea calculation considers the grid cell center of the topo and not the grid cell extent (perhaps by design?)

Including a buffer of +/- dx and dy around the x/y low/hi extent of the topo in the topoarea calculation addressed this issue for me. 

Is the use of topo grid cell centers by design or an error?

### DIFF
--- a/src/2d/shallow/topo_module.f90
+++ b/src/2d/shallow/topo_module.f90
@@ -1360,7 +1360,10 @@ recursive subroutine topoarea(x1,x2,y1,y2,m,area)
          ! only using coarsest topo grid -- compute directly...
          call intersection(indicator,area,xmlo,xmhi, &
              ymlo,ymhi, x1,x2,y1,y2, &
-             xlowtopo(mfid),xhitopo(mfid),ylowtopo(mfid),yhitopo(mfid))
+             xlowtopo(mfid)-dxtopo(mfid), &
+             xhitopo(mfid)+dxtopo(mfid), &
+             ylowtopo(mfid)-dytopo(mfid), &
+             yhitopo(mfid)+dytopo(mfid))
 
     else
         ! recursive call to compute area using one fewer topo grids:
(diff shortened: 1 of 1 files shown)


rjleveque:
Thanks for bringing this up @kbarnhart!

We view the topo values as being pointwise values at the centers of the topo cells, as explained to some extent in https://www.clawpack.org/grid_registration.html.

You are right that this means if the topo grid has exactly the same extent as the computational domain then the points don't extend to the boundaries and sometimes this keeps GeoClaw from running (it depends on what percentage of the domain is found to be outside the convex hull of the topo points).  We do this because a bilinear function is created interpolating at the topo points, which may not be valid outside the hull.

But we should revisit this, since it often happens that we download topo that seems to cover the same region as the domain and run into this issue, and then have to either download a slightly larger region or cut down the domain size by at least dx/2 or dy/2 on each side.  

We should probably check for this particular case and do something reasonable like use the bilinear from the boundary topo points to extrapolate, which is probably what it's doing anyway in cases where it doesn't die because the percentage is small enough.

I'll have to think some more about your proposed change to see if that's what this does?

mandli:
My 2¢ here:
- We need to adopt a consistent interpretation of the topography data inside GeoClaw and handle translations either at the point we read data, or as a pre-processing step, which we could add to `topotools`.
- I would also view a topography that exactly lies within every cell to not actually cover the domain due to interpolation issues, especially with the boundaries.
- We still can run into problems, as @mjberger found recently for a global problem and is related, but the fix was to change the extent of the data.
- If you have synthetic topography, I would highly recommend using the built-in topographical function evaluation capabilities if you can.

kbarnhart:
@rjleveque , @mandli - Thanks for your comments. I though a bit more about this and I came to the conclusion that this fix does not make sense so I'll close the PR. 

I think my mistake was considering that the topo extent perfectly covered the domain rather than the extent of the topo grid cell centers. I see that for the purposes of interpolation it is not entirely clear how to handle the area that is exterior to the outermost grid cell centers. 

In my application my topo and my mesh were perfectly aligned at the finest amr level so there was never a grid cell center that was outboard of defined topo. However, that is not guaranteed. 

And yes, I had specified my files to indicate registration using topotools. 

Thanks for helping me think this one through!

PR #640, opened 2025-03-18
Add vorticity plot helper functions

mandli:
Adds functions to compute and plot vorticity from GeoClaw.


rjleveque:
@mandli: Sorry I'm just getting to looking at this.  I notice you only compute vorticity on the interior points of each patch using central differencing, which may leave gaps at patch edges when plotted.  I have a version that uses one-sided differences at the edges if that's of interest.

I also found an old version of `valout.f` that computes and outputs the vorticity as a 5th column in the fort.q files, since it can be computed here using central differencing everywhere because the ghost cells are filled with data from neighboring patches.

Another alternative is that binary output dumps the full arrays, including ghost cells (which are missing in ascii output) and with a bit more work these can be preserved and used to calculate vorticity in setplot.

Let me know if any of these are of interest and we could work on it at the upcoming hackathon.

In the meantime, fine to merge this if you want it in the next release.

mandli:
I would say so on all fronts.  I think I briefly tried to do one-sided differences and it looked funny, especially at patch boundaries, but that may not be an issue for binary output.

PR #642, opened 2025-04-22
Switch to PyTest

mandli:
A clean attempt at switch the tests over to PyTest.  Also moves the tests out of a dedicated directory and aligns each with an example, modifying only what needs to be modified.


mandli:
The new `dtopo` example, based on the `dtopo1` test, is failing due to slight discrepancies in the gauges.

mandli:
Multilayer currently not compiling.  May be a lapack install issue?

mandli:
There are commits that should be merged in other repositories before this may work:
- clawpack/amrclaw#298
- clawpack/amrclaw#301
- clawpack/clawutil#177

rjleveque:
PR #649 suggests merging that on before this, but I think #649 has problems (see the discussion there), so perhaps the regression data needs to be updated in this PR?

mandli:
Yeah, we can update the regression data when necessary.

mandli:
I have done a review of all the tests and it appears that once a couple of bugs were fixed that the problems may now stem from the use of the modifications introduced in clawpack/riemann#183.  Below are comparisons of the old test data and the new.

multilayer_4
netcdf_bowl
particles_1
particles_2
vortex_0
vortex_1
vortex_2
vortex_3
vortex_4

bowl_slosh
chile_adjoint
dtopo
multilayer_1
multilayer_2
multilayer_3

mandli:
I also should add that I added an example/test to address the recent changes to the Riemann solver.  The test is in vortex and implements the moving vortex test case.  It currently is simply a regression/validation test, but we could add a convergence test to it.

The more pressing matter though is where to put this example.  I currently added a directory called `shallow` and stuck it in there as it has nothing to do with tsunamis or storm surge.  There are other examples in tsunami that also are not explicitly tied to tsunamis, or moving topography.  I am wondering what we would like to do about this.  I know I have gotten some questions about where some examples are and people did not think to look in tsunami if they were not doing tsunamis.

rjleveque:
Regarding where to put examples that aren't a tsunami, surge, etc.,  we could just put them in the top level of the examples directory, e.g. in `geoclaw/examples/vortex`. The `bowl-slosh` example might belong there too. Things like `bowl-radial` and `radial-ocean-island-fgmax` are more ambiguous, since these were developed with tsunami modeling in mind but test propagation and inundation more generally.

rjleveque:
@mandli: regarding your comment
> I have done a review of all the tests and it appears that once a couple of bugs were fixed that the problems may now stem from the use of the modifications introduced in https://github.com/clawpack/riemann/pull/183. Below are comparisons of the old test data and the new.

I'm not sure I understand the plots.  Several of them are labelled Gauge 1 and are very different from one another.  Also one of them has a discontinuity in q[3,:] at t=0.5.  What are these showing?

mandli:
> Regarding where to put examples that aren't a tsunami, surge, etc., we could just put them in the top level of the examples directory, e.g. in `geoclaw/examples/vortex`. The `bowl-slosh` example might belong there too. Things like `bowl-radial` and `radial-ocean-island-fgmax` are more ambiguous, since these were developed with tsunami modeling in mind but test propagation and inundation more generally.

Makes sense.  Will change over and propose some examples to move into an upper directory.

mandli:
I am going to close this as it has become too monolithic and conflicting with the current branch of GeoClaw.  A smaller PR with the essential changes to the `ClawpackTestRunner` will be issued based on some of what was done here, but with the goal of a minimal set of changes to get the tests working again.

PR #643, opened 2025-04-30
Bouss periodic and matrix reentrant corner and wall fixes

mjberger:
while implementing periodic bcs found bugs in wall bcs. Also fixed big bug in reentrant corrner. Also using new version of petsc with different options and modified Makefile in radial_flat. Also tested with examples from implicit paper.


rjleveque:
I've also tested this pretty well, so I am comfortable merging it in.

PR #645, opened 2025-05-03
Remove test output from new storm surge tests

mandli:
Addresses comment in #531


mandli:
Rest of tests will not work either but keeping them for the time being so that they can be modified to do so.

rjleveque:
Thanks @mandli!

PR #646, opened 2025-05-06
Trying to fix storm.py

rjleveque:
@mandli: I'm trying to run all the GeoClaw examples and  I don't know what's wrong with this.

I fixed nested double quotes and then it didn't like an f-string split between lines, but now it still gives an error...

```
$ cd $CLAW/geoclaw/examples/storm-surge/ike
$ make data
            
  File "/Users/rjl/git/clawpack/geoclaw/src/python/geoclaw/surge/storm.py", line 1070, in write_geoclaw
    f"{np.datetime_as_string(self.time_offset,unit='s')}\n\n")
TypeError: input must have type NumPy datetime
```

### DIFF
--- a/src/python/geoclaw/surge/storm.py
+++ b/src/python/geoclaw/surge/storm.py
@@ -1065,7 +1065,7 @@ class Storm(object):
             with open(path, "w") as data_file:
                 # Write header
                 data_file.write(f"{num_casts}\n")
-                if isinstance(self.time_offset, datetime.datetime):
+                if isinstance(self.time_offset, np.datetime64):
                     data_file.write(\
                       f"{np.datetime_as_string(self.time_offset,unit='s')}\n\n")
                 else:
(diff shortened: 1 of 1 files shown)


mandli:
Looking into this.  My guess is that something is in the wrong format, although I am not sure how it was not caught (testing).

mandli:
It was a bug, the type check
```python
if isinstance(self.time_offset, datetime.datetime):
```
should be
```python
if isinstance(self.time_offset, np.datetime64):
```
so that the code is only run if that `self.time_offset` is a `numpy.datetime64` object.

mjberger:
while you're looking at input data, I'm having a problem with the range stuff in setrun.py making data files with np.int or np.float with the number, instead of just the number  

anyone else have this problem?

 — Marsha

rjleveque:
Thanks @mandli, now it's working.

PR #647, opened 2025-05-17
Add support for overriding topography array ordering

mandli:
This PR adds the abiility to override the priority that a topography file receives.  

Previously topography files are ordered first by their resolution and then by the order they are given in `setrun.py`.  This PR adds a parameter `override_order` so that the order that the topography files are specified in `setrun.py` always dictates the order that they will be used.

This is useful if you have two or more topography files that are nearly the same resolution and there is a preference for their order that does not match this.

### DIFF
--- a/src/2d/shallow/topo_module.f90
+++ b/src/2d/shallow/topo_module.f90
@@ -225,6 +225,9 @@ contains
                 ! The finest topography will be given priority in any region
                 ! mtopoorder(rank) = i means that i'th topography file has rank rank,
                 ! where the file with rank=1 is the finest and considered first.
+                !
+                ! If override_topo_order is .true. then the order that the files were
+                ! specified is used directly
                 if (override_topo_order) then
                     do i=1, mtopofiles
                         mtopoorder(i) = i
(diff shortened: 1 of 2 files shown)


rjleveque:
Thanks @mandli, this seems to work fine.

Notes:

To use this, set 
```
    rundata.topo_data.override_order = True
```
in `setrun.py`, in which case the first topofile listed is given highest priority, with decreasing priority for each subsequent topofile.  This means that the topo value in a grid cell is computed by first computing it using the *last* topofile listed, and then working backwards to update the value by removing the portion for the area of the cell covered by the previous topofiles and replacing it by the value computed on the intersection of the cell with that topofile.

Note that if 
```
    rundata.topo_data.override_order = False
```
(the default value if you do not set it) then the original standard algorithm is used, with higher resolution topofiles (smaller area based on `dxtopo*dytopo`) having higher priority.   If two topofiles have exactly the same resolution, then the one listed **last** has priority.  Note that this is different from what might be expected based on the new `override_order` parameter!

mandli:
@rjleveque that makes sense.  I will implement the changes.

rjleveque:
Looks good, and I'm working on some new documentation to appear at https://www.clawpack.org/topo_order.html

PR #648, opened 2025-05-19
error when nw==11

dlgeorge:
was reaching an error in dclaw from parse_values with 8 fgout variables (size(word)==10), but I think error/stop was premature (?). Probably double check that change is okay.

### DIFF
--- a/src/2d/shallow/utility_module.f90
+++ b/src/2d/shallow/utility_module.f90
@@ -84,13 +84,14 @@ contains
            endif
 
         nw = nw + 1
-        word(nw) = trim(adjustl(str2(1:pos2-1)))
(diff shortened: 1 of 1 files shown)


mandli:
Huh, this is just a straight up bug, but do we also want to just increase the number of words possible on a line?

dlgeorge:
adding more words --up to you, 10 turns out to be just enough for dclaw fgout (1-8 currently) but q/meqn seems to have a way of growing in dclaw, so maybe for the future. I  have a branch that I increased to 12...

mandli:
How about 16 to make it a nice power of 2 and confuse us as to why we may have chosen such a specific number?

dlgeorge:
ok, sounds good. I can issue pull request with 16.

PR #649, opened 2025-05-21
Change Timing for Gauge Updating

mandli:
This PR moves the `update_gauge` call to after a step is taken.  It also puts a condition on an initial time `abs(t0 - time) < TOLERANCE` so that the initial time is still recorded at the initial time.


mandli:
This seems to work although there are two questions:
- Does the impact the `min_time_output` functionality for gauges?
- Should we not record the gauge secondarily after the first step?

My guess is no to both, but it would be good to have a double check.

This should be merged before #642.

PR #651, opened 2025-05-30
fix timing_line definition in valout.f90

rjleveque:
Add a missing comma that didn't seem to cause a problem with earlier gfortran but throws an error with 15.1.0


rjleveque:
The same fix is needed in amrclaw...

PR #654, opened 2025-06-04
post 5.11 fixes for dclaw: fgout_tools and fgout xarray backends 

kbarnhart:
@rjleveque - A few v5.11 changes made using the fgout tools with D-Claw output not work as expected because of how the default q_out_vars are handled. 

I didn't find this issue until now because it took me a while to have a need to re-process pre-5.11 output with post 5.11 code.

Happy to revise further if you think this needs additional updates.

Thanks!

### DIFF
--- a/src/python/geoclaw/fgout_tools.py
+++ b/src/python/geoclaw/fgout_tools.py
@@ -342,11 +342,13 @@ class FGoutGrid(object):
         self.output_format = output_format
 
         # list of which components to print:
-        if 'eta' in self.qmap.keys():
-            # default: h,hu,hv,eta as in previous versions of GeoClaw
-            self.q_out_vars = [1,2,3,self.qmap['eta']]
+        if qmap == 'dclaw':
+            # default: h,hu,hv,hm,pb,hchi,bdif,eta, DClaw
+            self.q_out_vars = [1,2,3,4,5,6,7]
         else:
-            self.q_out_vars = [1,2,3]  # if qmap['eta'] not defined
+            self.q_out_vars = [1,2,3] # use Geoclaw values as default.
+        if 'eta' in self.qmap.keys():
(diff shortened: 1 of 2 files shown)


rjleveque:
@kbarnhart: I think this is fine but just to be clear we're on the same page, I think it shouldn't be needed for reading post-511 geoclaw or dclaw data, since the `fgout_grids.data` should then include a line that lists the `q_out_vars` and this is used in `read_fgout_grids_data`, so whatever is set in `FGoutGrid.__init__` gets overwritten.  

But `read_fgout_grids_data_pre511` does not read this in, since it wasn't user-setable pre-511, so for reading older data you do need your fix.

Does this agree with what you found?

kbarnhart:
@rjleveque 

You are correct that I need this only to read pre-v5.11 output with new code. 

The switch to `read_fgout_grids_data_pre511` is to address this error:

```bash
File "clawpack/geoclaw/src/python/geoclaw/xarray_backends.py", line 224, in open_dataset
    fgout_grid.read_fgout_grids_data()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "clawpack/geoclaw/src/python/geoclaw/fgout_tools.py", line 500, in read_fgout_grids_data
    assert next_value_int, err_msg
           ^^^^^^^^^^^^^^
AssertionError:
*** Expecting integer output_style next in fgout_grids.data
    If this is fgout data from before v5.11, try using
    read_fgout_grids_data_pre511
```

Switching to use `read_fgout_grids_data_pre511` if the try-except block fails results an almost-functional fgout object. However, without the second fix, I don't have access to the dclaw variables beyond h, hu, and hv. 

Hopefully that explains all the changes.

rjleveque:
@kbarnhart : Sounds good, thanks for fixing this!

PR #655, opened 2025-06-05
clean up when to call resetBouss instead of every step, adding block …

mjberger:
…coo option, adding petsc options to output timing file

### DIFF
--- a/src/2d/bouss/Makefile.bouss
+++ b/src/2d/bouss/Makefile.bouss
@@ -159,6 +159,7 @@ COMMON_SOURCES += \
   $(BOUSSLIB)/buildSparseMatrixSGNcrs.f90 \
   $(BOUSSLIB)/prepBuildSparseMatrixSGNcoo.f90 \
   $(BOUSSLIB)/buildSparseMatrixSGNcoo.f90 \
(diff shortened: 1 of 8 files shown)


rjleveque:
@mjberger: I think you forgot to include `buildSparseMatrixSGNcoo_blocks.f90` so I can't test it.

Some minor suggestions:

1. You introduced `is_finalized`, maybe it would be clearer down the road to call this `bouss_finalized`, in keeping with `topo_finalized` and `aux_finalized`?

2. In `bouss_module` maybe change the comments in these lines:
```
     !crs = .false.  ! uncomment this line to force COO with SGN for testing
     !origCooFormat = .false.  ! using block format with coo
```
to
```
     !crs = .false.  ! set to .false. to force COO with SGN for testing
     !origCooFormat = .false.  ! set to .false. to use block format with coo
```
to be less confusing, particularly if the user uncomments a line and then changes `.false.` to `.true.` in the command without updating the comment.

3. It might be helpful to have a few more comments somewhere about the differences between these different formats and why they are all available. (For someone else debugging down the road.)

mjberger:
new comments in bouss_module

PR #665, opened 2025-08-21
add kmltools.dtopo_contours2kmz

rjleveque:
This function makes a kmz file with contours of dtopo vertical deformations for one or more dtopo files.

Used for example to make the `buried_dtopo_contours.kmz` file posted on [this webpage](https://depts.washington.edu/ptha/CopesHubTsunamis/dtopo/CSZ_groundmotions/animations/), which can also be [viewed on earth.google.com](https://earth.google.com/earth/d/1vpzrgZ6dKG4GlMNXpqA3qSenj8uCrK4g?usp=sharing) (open the menu on the left to toggle different events).  For reasons I don't understand, it shows up at lower resolution on the web version than on the desktop.


mandli:
This may be a future thing to do, but the imports and functionality in this file needs to be reworked.  For instance `matplotlib` is imported in a few places in a bunch of different ways.  Same goes for `numpy`.

In any case, if it works maybe we should merge it in?  I just worry that we will never get back to this and refactor it unless we do it know.

PR #669, opened 2025-08-30
updated for override_order line in topo.data affecting 1d radial flat…

naveenragur:
While reading topo file name, the previous line was read. Updated to match topo_module.f90 from 2d/shallow.


rjleveque:
@naveenragur: Thanks for catching this!  It also affects all the examples in `examples/1d_classic`.

Currently in 1D we only allow a single topofile, so the ordering doesn't matter, but since the parameter is written out we need to read it.  In the future perhaps we will also implement multiple topofiles in 1D, in which case it could be more useful.

Issue #667, opened 2025-08-27
Pass `actualstep` to `b4step2`


mandli:
When playing with some debugging runs I found that `actualstep` is not being passed to `b4step` correctly.  It used to be passed to `stepgrid` and then on to `b4step`, but moving `b4step` to `advanc` makes that not work anymore.  I think I have a fix, but this does mean `acutalstep` is no longer available in `stepgrid`.

mjberger:
I don't think the actualstep thing is  problem. b4step call might be.

The only thing b4step2 does if actualstep is true is update topo if the aux arrays aren't yet finalized. This does not happen if  actualstep is false. So the value of actualstep isn't an issue.

The call to b4step2 might be an issue though if the other things it does still need to be done, such as setting storm fields?   Is this needed, or is it created in prepbigstep already by calling setaux and then coarsening?    The other stuff, like checking for nans and speed limit checking might not be relevant for 1 step of prepbigstep taken for error estimation? 

I would put a note in prepbigstep that says to think about putting in a call to b4step2 before calling stepgrid if it's actually needed, but it doesn't appear to be.

rjleveque:
It seems like a problem that the call to `b4step2` in `advanc` has the wrong number of parameters, since it's missing `actualstep`.  Shouldn't `actualstep = .true.` in this call since this is done jub above where `stepgrid` is called with `actualstep = .true.`?

@mandli: Why do you say `actualstep` won't be available in `stepgrid` since it still seems to be in the call to `stepgrid` from `advanc` and `prepbigstep`?

mandli:
> It seems like a problem that the call to `b4step2` in `advanc` has the wrong number of parameters, since it's missing `actualstep`. Shouldn't `actualstep = .true.` in this call since this is done jub above where `stepgrid` is called with `actualstep = .true.`?

Yup, `actualstep == .true.` when called from `stepgrid` so from `advanc` I belive it should have `actualstep == .true.` and when called from `prepbigstep` `actualstep == .false.`.  

> [@mandli](https://github.com/mandli): Why do you say `actualstep` won't be available in `stepgrid` since it still seems to be in the call to `stepgrid` from `advanc` and `prepbigstep`?

`stepgrid` does not use `actualstep` save for calling `b4step2` so I figured it was best removed, but this is a small point.

[...]

PR #670, opened 2025-09-05
Fix missing `actualstep` in `b4step` calls

mandli:
Addresses #667

### DIFF
--- a/examples/multi-layer/bowl-radial/setrun.py
+++ b/examples/multi-layer/bowl-radial/setrun.py
@@ -87,8 +87,8 @@ def setrun(claw_pkg='geoclaw'):
     # Index of aux array corresponding to capacity function, if there is one:
     clawdata.capa_index = 0
 
-    
-    
+
+
     # -------------
     # Initial time:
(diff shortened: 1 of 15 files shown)


mandli:
This also adds `actualstep` checks for storm fields and multilayer stuff.

mandli:
Also adds an initialization for `tend_force_dry` as it is used without being defined in `qinit_module` otherwise.  Sets the default value to `tend_force_dry = -huge(1.d0)`.

rjleveque:
Lots of white space is cleaned up too, which is good to do but makes it a bit hard to see all the actual code changes.

One thing I noticed is that the call to `stepgrid` in `advanc.f` and `prepbigstep` no longer has `actualstep` in the parameter list, but it is still in calling sequence in `stepgrid.f`.  It's not used in `stepgrid` now, so remove it, and also from `multilayer/stepgrid.f` and `bouss/stepgrid.f` ?

rjleveque:
Seems to work, but this time I will run all the examples again before releasing!

PR #672, opened 2025-09-07
change dtopotools._load_sift_unit_sources to use pooch

rjleveque:
Try to download directly from `http://sift.pmel.noaa.gov/ComMIT/compressed/info_sz.dat` if it's not in the pooch cache.

Also updated the version in the data/ directory (no change in unit source data itself) and this will still be used if pooch fails. `known_hash` can be used in future to see if file has changed online. 

This change was motivated by the fact that `pip install clawpack` does not expose the local version in data/ and so using `dtopotools.SiftFault` fails in Github Actions when building a Juptyer Book there.


mandli:
This looks fine to me, but I had to ask if "pooch" is an acronym or just an entertaining name for a data repository.  Is "dog house" involved?

rjleveque:
And for the particular case of the file from http://sift.pmel.noaa.gov/ComMIT/compressed/info_sz.dat,  I like the fact that it tests a checksum and can alert us if NOAA changes their unit source definitions at some point in the future, so we don't have a stale version hardwired into GeoClaw.

PR #673, opened 2025-09-08
suppress plot made by make_shoreline_xy

rjleveque:
`topotools.Topography.make_shoreline_xy` makes a contour plot only to extract the contour lines.
Close the figure so it doesn't appear when not wanted in a notebook.


rjleveque:
Yes, I have run into cases when making animations in particular where it runs into memory problems if the figures aren't closed properly. I've never had to delete a figure after closing.

As far as I can see for this PR `close` is enough, and behaves as desired, so I'll go ahead and merge.

Issue #656, opened 2025-06-25
Potential Confusion Due to Format Specifier Inconsistency


hydrocoast:
I noticed that the numeric values used to specify the binary format are handled differently for gauge output compared to other outputs.
This may be a minor inconsistency and not something that necessarily needs to be fixed, but I thought it could potentially cause unexpected errors during development.

In gauges_module.f90, the file_format specifier is:
1: ascii, 2: binary64, 3: binary32.

On the other hand, in valout.f90 and fgout_module.f90, the output_format specifier is:
1: ascii, 2: binary32, 3: binary64 — meaning that 2 and 3 are reversed relative to the gauge output.

https://github.com/clawpack/geoclaw/blob/162d2c2fef81b083a2f638272d91dbb534012011/src/2d/shallow/gauges_module.f90#L292-L301

https://github.com/clawpack/geoclaw/blob/162d2c2fef81b083a2f638272d91dbb534012011/src/2d/shallow/valout.f90#L405-L413

https://github.com/clawpack/geoclaw/blob/162d2c2fef81b083a2f638272d91dbb534012011/src/2d/shallow/fgout_module.f90#L592-L602

mandli:
Maybe we consider the gauges a bug and fix it so it matches?  If we put it in the release notes at least someone may be warned, or we could put an explicit warning in the Python.

I have been working on a lot of NetCDF reading and writing lately and may be in a state that I could get something working.  Maybe I will play with it a bit.

mandli:
This has now been addressed and released.  Thanks @hydrocoast!

PR #675, opened 2025-09-08
Switch gauge binary file format order

mandli:
Addresses #656.

### DIFF
--- a/src/2d/shallow/gauges_module.f90
+++ b/src/2d/shallow/gauges_module.f90
@@ -749,10 +749,10 @@ contains
                       (gauges(gauge_num)%data(k, j), k=1,nvals)
             end do
 
-        else if (gauges(gauge_num)%file_format >= 2) then
+        else if (gauges(gauge_num)%file_format == 2 .or.    &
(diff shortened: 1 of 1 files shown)


rjleveque:
I'm confused at the moment.  With this new version, if I set

```
   rundata.gaugedata.file_format = 'binary32'
```
in `setrun.py` and do `make data` then `gauges.data` contains:

```
# File format (1=ascii, 2=binary64, 3=binary32)
3 
```

which is the old value 3 rather than 2.  This makes sense since this file is written by `amrclaw/src/python/amrclaw/data.py` which hasn't been modified and probably needs to be.

But then when I run the code the resulting `gaugeNNNNN.txt` file has a header that contains
```
# file format binary32, time series in .bin file
```
even though `gauges_module.f90` was modified, so shouldn't it be writing out `binary64` since it reads `file_format == 3` from `gauges.data`?  Also the resulting `bin` file is smaller by a factor 2 than what I get when I specify `binary64`, so it seems to do what's desired but I don't understand why.

rjleveque:
Looks good now, thanks!

PR #682, opened 2025-12-27
Center points updates

rjleveque:
Fix a bug and improve `center_points.py`. Use to center gauges or fgmax, fgout points in computational cells.

Also added utility function `center_gauges.py` to center gauges if the desired locations are already in a `setrun.py` function.


rjleveque:
@mandli: Good point, perhaps `center_points` should be in `util.py` since it can be used also for fgmax or fgout points, and `center_gauges` in `gauges.py`.

PR #686, opened 2026-01-12
Add make_function method to Topography and DTopography objects

rjleveque:
Make it easier to read in a topo or dtopo file and define an interpolating function that can be used e.g. to plot a transect of the topo or dtopo.

WIP... example to appear.


rjleveque:
I put a simple topo example here:
http://depts.washington.edu/clawpack/samples/topo_func_demo.html
to be added to the apps repository eventually along with a dtopo example.

This seems to work fine, so I'm going to merge it.

PR #687, opened 2026-01-13
Fix topo_module.f90 so that override_topo_order works as desired

rjleveque:
The earlier change in commit 32dcc947 changed the order `mtopoorder(i)` was set but still set the value to `i`, whereas it should be `mtopofiles + 1 - i` in order for topofiles to be used preferentially with the first file listed having lowest priority.

The old version was in v5.13.0 and v5.13.1 but should be fixed in v5.14.0.


mandli:
I thought I had tested this to make sure that it was correct before merging in that PR.  I am confused though, doesn't this change do the same thing as what was being done?

PR #688, opened 2026-01-15
Draft: Add example of color in x-y plots

kbarnhart:
Related to https://github.com/clawpack/visclaw/pull/321

### DIFF
--- a/examples/tsunami/bowl-radial/setplot.py
+++ b/examples/tsunami/bowl-radial/setplot.py
@@ -351,16 +351,19 @@ def setplot(plotdata=None):
     plotaxes.title = 'Scatter plot of surface'
 
     def r_surface_and_vel(current_data):
-        from numpy import sqrt, nan
+        from numpy import sqrt, nan, divide, zeros
         x = current_data.x
         y = current_data.y
         h = current_data.q[0,:,:]
-        u = current_data.q[1,:,:]/h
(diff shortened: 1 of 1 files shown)


rjleveque:
Thanks @kbarnhart, nice example.

It works fine, but does throw a warning about divide by zeros (later overwritten by nan's).  If you want to avoid this, you could use:

```
    def r_surface_and_vel(current_data):
        from numpy import sqrt, nan, divide, zeros
        x = current_data.x
        y = current_data.y
        h = current_data.q[0,:,:]
        dry_tol = 1e-3
        u = divide(current_data.q[1,:,:], h,
                   where=h>dry_tol, out=zeros(h.shape))
        v = divide(current_data.q[2,:,:], h,
                   where=h>dry_tol, out=zeros(h.shape))
        eta = current_data.q[3,:,:]
        vel = sqrt(u**2 + v**2)
        vel[h<dry_tol] = nan
        eta[h<dry_tol] = nan
        r = sqrt(x**2 + y**2)
        return r,eta,vel
```

This sets `u,v` to 0 where `h <= dry_tol`, which is generally what's wanted instead of `inf`.
Or you could use `out=nan*zeros(h.shape)` in `divide()` to set `u,v` to `nan` more directly whereever `h <= dry_tol`.

PR #692, opened 2026-02-09
fix bug in topotools.plot call to pcolorcells

rjleveque:
parameter `axes` passed into `plot` is `ax` in call to `pcolorcells`


mandli:
That would explain a couple of things.

PR #693, opened 2026-02-14
Fix dt passed uninitialized to b4step2

carlosmunozmoncayo:
Fix uninitialized `dt `passed to `b4step2`  in `advanc`.


rjleveque:
Thanks @carlosmunozmoncayo, good catch!  We don't use `dt` in the default version `src/2d/shallow/b4step2.f90` but good to get this fixed for users who need it.

mandli:
`b4step` was broken, but the new code we added when we moved `b4step` to `advanc` needed to use `delt`, rather than what it was when being called from `stepgrid`.  You are right though, any issues would have been present already.

PR #695, opened 2026-02-27
Fix NetCDF coordinate reads in topo_module: use varid instead of dimid

mandli:
`topo_module.f90` retrieves coordinate data using:
```fortran
nf90_get_var(nc_file, x_dim_id, ...)
```
where `x_dim_id`/`y_dim_id` are dimension IDs and not variable IDs.  The NetCDF Fortran API requires variable ID for `nf90_get_var`.

Because NetCDF IDs are integers, this sometimes silently works when a dimid numerically matches a varid, but fails for valid CF files (e.g., GEBCO NetCDF4) with:
```
NetCDF: Start+count exceeds dimension bound  (-57)
```
This PR simply switches to the proper variable query
```fortran
nf90_inq_varid(nc_file, trim(x_dim_name), x_var_id)
nf90_inq_varid(nc_file, trim(y_dim_name), y_var_id)
```
and checks those coordinates.

Also adds a small fix for nearest-index searches with
```fortran
minloc(abs(xlocs - xll), dim=1)
```

### DIFF
--- a/src/2d/shallow/topo_module.f90
+++ b/src/2d/shallow/topo_module.f90
@@ -634,7 +634,7 @@ contains
                 
                 ! get x and y dimension info
                 call get_dim_info(nc_file, num_dims_tot, x_dim_id, x_dim_name, &
-                    mx_tot, y_dim_id, y_dim_name, my_tot)
+                     x_var_id, mx_tot, y_dim_id, y_dim_name, y_var_id, my_tot)
                     
                 allocate(xlocs(mx_tot),ylocs(my_tot))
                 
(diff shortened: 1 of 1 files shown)


mandli:
This now works for the GEBCO data I was testing on, but I want to change over all the NetCDF places I think this is happening while I am at it.

mandli:
There are some changes to the `util.wrap_coords` function for topography that this was revealed by, but I think that will go into another PR to keep this cleaner.

mandli:
This has been merged into #701

PR #696, opened 2026-03-04
Add more robust wrap_coords function

mandli:
This rewrites and makes more general the way that the `wrap_coords` function was working including more general wrapping options.


mandli:
This has been merged in with #701

PR #697, opened 2026-03-17
Switch to Using ClawpackTestRunner for Testing

mandli:
This PR aims to use the new testing framework now used in [AMRClaw](https://github.com/clawpack/amrclaw) and [Classic](https://github.com/clawpack/classic) to run regression tests from [clawutil](https://github.com/clawpack/clawutil)'s `ClawpackTestRunner` class.  This switches everything over to using [PyTest](https://docs.pytest.org/) and as much functionality as possible from there.

Major goals for this particular PR is to use the same testing structure as was in AMRClaw and Classic, but change things minimally to get things working before moving onto to larger structural changes.


mandli:
Phew, all the tests from before are now implemented!  There are a few issues that need to be looked at though:

1. Tests marked as `xfail`/`xpass`:
    - [x] tests/test_storm.py::test_storm_io[ibtracs] - This is expected as ibtracs have not been fully implemented
    - [ ] tests/test_dtopotools.py::test_dtopo_io - Failing occasionally, need to look at read/write precision
    - [ ] tests/test_data.py::test_read_fgmax_data_multiple_grids - FGmaxData.read does not yet round-trip this multi-grid/point_style case correctly.
correctly.
    - [ ] tests/test_data.py::test_dtopo_data_roundtrip - DTopoData.read currently fails to parse this written dtopo.data round-trip case.
    - [ ] examples/tsunami/island-particles/test_particles.py - Slight gauge mismatch
    - [ ] examples/tsunami/dtopo/test_dtopo.py::test_dtopo - Slight gauge mismatch
    - [ ] examples/tsunami/chile2010_adjoint/test_chile2010_adjoint_forward.py - Slight gauge mismatch
2. Skipped
    - [ ] tests/test_dtopotools.py::test_geometry - Uses old tools and may be duplicative with use of newer tools
    - [ ] tests/test_dtopotools.py - Skipped comparison with old tools

More test coverage is possible, but if we can merge this in, we can also systematically correct the remaining regression data issues.

mandli:
Current markers implemented (in pytest.ini):
- Test type (choose one):
    - python: pure Python unit tests
    - regression: example-based regression tests (Fortran)
- Execution characteristics:
    - slow: slower tests not intended for the default quick CI path
    - remote: tests that require fetching a file from the web
    - netcdf: tests requiring NetCDF support
- Domain / feature grouping:
    - tsunami: tsunami related tests
    - storm: storm surge related tests
    - multilayer: multilayer shallow water related tests
    - adjoint: inverse/adjoint solver releated tests
    - adjoint_forward: forward runs that use inverse/adjoint solver, uses adjoint data

mandli:
I have now split the testing into three groups:
1. `python` unit testing, which will eventually also include linting.
2. `regression` testing
3. `slow` regression testing

There's really only one slow regression test right now, and it really does not do anything but run a longer run.  The intent is that the slow regression testing will happen periodically (daily at 6 AM UTC right now).  We could also add another one that would do the gallery examples, or do the Jupyter notebooks associated with the repos.  Wanted to get an example running though so that we can think about how to use it.

If this looks good, I can move this structure to AMRClaw and classic when it makes sense.

rjleveque:
Merging this along with https://github.com/clawpack/classic/pull/96 and https://github.com/clawpack/amrclaw/pull/310.

PR #700, opened 2026-04-07
fix some character length issues in 1d topo_module.f90

rjleveque:
input variable strings should have length *

code was crashing for me with garbage in strings that trim wasn't removing.


mandli:
Missed this when I changed over a bunch of the path variables.  Looks good to me.

mandli:
This does remind me that we need to test the 1D code.

PR #716, opened 2026-07-28
Add better OWI format file support

mandli:
This is pulling in better support for OWI file reading and writing that existed in another branch and previously implemented.  This PR brings in the new functionality provided and matches any improvements available with the existing code.


mandli:
This also makes a note about the OWI field pressure precision.  Should not be an issue in most cases, but documented now.

PR #718, opened 2026-07-29
add more comments to examples/bouss/setenv.sh

rjleveque:
To make it clearer that mpif90 path should be consistent with mpiexec.

@mjberger: I added the petsc path to mpif90 (commented out) to the example, is this correct?


mjberger:
yes - is correct. works in Makefile too

PR #720, opened 2026-07-29
Speed up and reduce the cost of the testing CI

mandli:
# Motivation

The regression CI had grown slow and was consuming a large share of GitHub Actions minutes. An audit confirmed two root causes:

1. Redundant Fortran rebuilds dominate regression wall time. Each regression test called build_executable(), which defaults to make new — a full clobber-and-recompile of ~140 Fortran sources into a shared in-tree object location, so no build was ever reused. The hot examples rebuilt the same binary once per parametrization (chile2010 ×6, bowl-slosh-netcdf ×6, met_forcing ×6) even though only a handful of distinct binaries exist.
2. macOS ran on every PR at the 10× minute multiplier, with no concurrency cancellation, no timeouts, and no caching — so superseded PR pushes stacked full 8-job runs to completion.

# Changes

Per-PR workflow (testing.yml) — faster and cheaper
- Added concurrency + cancel-in-progress so new pushes cancel stale runs.
- Added timeout-minutes to every job as a runaway backstop.
- Moved all macOS cells off the per-PR path (now 4 Linux jobs, was 8 with 4 macOS-10×).
- Added and not remote to the selections so per-PR CI makes no live network fetches; pinned clawpack/clawpack ref: master; added a paths-ignore filter for docs-only changes.
- Set up Fortran explicitly for the python-tests job (installing the clawpack meta-package compiles Fortran extensions, so this no longer relies on the runner image shipping gfortran).

**Nightly workflow (slow-tests.yml → "Nightly and slow tests") Changes**

Per-PR workflow (testing.yml) — faster and cheaper
- Added concurrency + cancel-in-progress so new pushes cancel stale runs.
- Added timeout-minutes to every job as a runaway backstop.
- Moved all macOS cells off the per-PR path (now 4 Linux jobs, was 8 with 4 macOS-10×).
- Added and not remote to the selections so per-PR CI makes no live network fetches; pinned clawpack/clawpack ref: master; added a paths-ignore filter for docs-only changes.
- Set up Fortran explicitly for the python-tests job (installing the clawpack meta-package compiles Fortran extensions, so this no longer relies on the runner image shipping gfortran).

**Nightly workflow (slow-tests.yml → "Nightly and slow tests") — preserves all coverage**
- New regression-tests and python-tests jobs carry the macOS matrix and the remote-marked tests moved off the per-PR path, so nothing loses coverage — it just runs once a day where the 10× cost is paid once.
- Added permissions: contents: read and timeout-minutes.

**Build reuse in the regression tests (~18 rebuilds → 5)**
- Introduced module-scoped fixtures that build each distinct xgeoclaw once and copy it per test (following the existing isaac pattern), replacing per-parametrization rebuilds:
  - bowl-slosh-netcdf: 6 → 1 (one NetCDF-enabled binary).
  - met_forcing: ~6 → 2 (plain_xgeoclaw / netcdf_xgeoclaw).
  - chile2010: 6 → 2 (fixtures pulled lazily so standard only builds plain and NetCDF variants only netcdf; the netcdf fixture uses the same FFLAGS the per-test build used).

# Housekeeping
- Removed stale .travis.yml (Python 3.6, git://, nose).
- Added .github/dependabot.yml for the github-actions ecosystem.
- Dropped the unused flake8 install (the lint step is commented out).

# Verification
- Both workflows are green on the branch: per-PR (4 Linux jobs) and a manually-dispatched nightly with all 7 jobs passing, including macOS gcc15 debug (-fcheck=all) and opt.
- The build-reuse change was verified against the golden regression data (locally and in CI). Because the shared binaries are produced by the identical make new invocation — just once and copied — results are unchanged; a NetCDF variant accidentally running on a plain binary would fail its gauge/aux check, and all pass under both debug and opt.


mandli:
Testing for slow-tests and macos runs now.  This also will eventually include more reusable build products to reduce the cost even further.

PR #725, opened 2026-08-01
Add a substring check for Fortran compiler for LAPACK inclusion

mandli:
Match on a substring of FC so versioned/pathed compiler names also work (e.g. setup-fortran >= 1.9.2 exports FC=gfortran-14 rather than gfortran).


mandli:
This should address the failure in #723.

PR #730, opened 2026-08-23
Add Development Directory `dev`

mandli:
Adding a `dev` directory to geoclaw so that it can hold the usual things that the `dev` directories do, but also as a tracked place for design documents such as roadmaps, refactor design records, and implementation plans that are meant to be in parallel and reviewed alongside commits and PRs when needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,7 @@ __pycache__
 
 # macOS
 *.DS_Store
+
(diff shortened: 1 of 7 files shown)


mandli:
This also adds some development plans that I lost since it was not being tracked.  This seems to be the suggested best-practice, especially when working with LLM assistants.

rjleveque:
Good idea!

PR #731, opened 2026-08-26
Remove vestigial line in gauges_module.mod

kbarnhart:
The following import in the gauges_module is not used. Can we delete it?

**The backstory:** The recent renaming of all the meteorological forcing files breaks D-Claw because there are a few places where D-Claw relies on parts of GeoClaw that rely on these files. At present we don't use any of the met forcing functionality (maybe in the future...) 

We can fully remove most of the interdependency, EXCEPT for this line.


kbarnhart:
Tagging @rjleveque @mandli @dlgeorge

mandli:
Probably a vestigial leftover from another way that I had things recording aux fields in gauges.

Larger conversation is about where you are running into issues with the rename and file dependencies.  Maybe another issue would be best to start for that though.

PR #732, opened 2026-08-26
Draft: Small ginit change

kbarnhart:
In current D-Claw work we have recently found that this change is needed to allow values to persist in aux across regrinding. 

For the moment we have made this change in a D-Claw specific ginit. But if the change does not cause problems in geoclaw, perhaps it should live here. For reference, in D-Claw we are working to create a storage layer of static material to put h<drytol. We'd like this to be in aux. If this change is not made, that value of aux is set to zero every time regridding occurs. With this change we can know which parts that aux variable need to be set. 

@rjleveque @mandli @dlgeorge

### DIFF
--- a/src/2d/shallow/ginit.f
+++ b/src/2d/shallow/ginit.f
@@ -36,6 +36,11 @@ c :::::::::::::::::::::::::::::::::::::::;::::::::::::::::::::
                 locaux              = igetsp(mitot*mjtot*naux)
                 ! do k = 1, mitot*mjtot*naux, naux  ! only set first component of aux to signal
                 do k = 1, mitot*mjtot*naux  ! KRB 2026/08/26 - set all components to NEEDS_TO_BE_SET rather than just aux(1)
+
+                ! NEEDS_TO_BE_SET must be consistently used across multiple routines: 
+                ! setaux: To not recompute (or persist) across gridding setaux must check 
+                !         NEEDS_TO_BE_SET before writing.  
+
                    alloc(locaux+k-1) = NEEDS_TO_BE_SET ! new system checks this val before setting
                 end do
                 
--- a/src/2d/shallow/ginit.f
+++ b/src/2d/shallow/ginit.f
@@ -34,7 +34,8 @@ c :::::::::::::::::::::::::::::::::::::::;::::::::::::::::::::
               node(store1,mptr)   = loc
               if (naux .gt. 0) then
                 locaux              = igetsp(mitot*mjtot*naux)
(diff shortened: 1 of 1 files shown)


mandli:
Thanks @kbarnhart.  I remember you bringing this up and being confused about why this might be needed.  I was able I think to puzzle out what I was not understanding and have a couple of comments:
- **The change should be fine in GeoClaw.** `setaux` used `NEEDS_TO_BE_SET` only on the topo component and is used mostly so that topo does not need to be recomputed if an existing grid at the same level overlaps with requested block.  Setting the rest won't matter unless they have modified `setaux` to look for `NEEDS_TO_BE_SET`.  This may take time though to set the arrays to this value rather than just assuming it though.
- **Future aux values may need this feature.** We could need this for aux components that are either expensive to compute or, as in your case, it needs to persist.  We should make it clear though that if we set `NEEDS_TO_BE_SET` that this value does not mean it has been, i.e. we may need to ensure that the logic holds out.
- **This alone will not solve the persistence problem in D-Claw.**  As I think you may have realized, the subroutines that set aux are `fillval` + `setaux`.  I assume that in D-Claw you have gotten the logic already so that this is solved, and you are simply copying in the values you need for `NEEDS_TO_BE_SET` but note that the idea behind this again was to reduce computational overhead rather than implement persistence.  Not that we cannot do this, but it was originally intended for something else.

I have not looked through what you did in D-Claw so far, but the implementation that I think should work would require (a) `setaux` to check for `NEEDS_TO_BE_SET` before writing and (b) `filval` to flag that component before `icall` so only un-copied cells get recomputed.

We may also want to add a comment mentioning what `NEEDS_TO_BE_SET` means, matching [`filval.f90:136`](https://github.com/clawpack/geoclaw/blob/7ef7a8162c6466b58f416736ae663594dcb7bd54/src/2d/shallow/filval.f90#L136) for consistency between the pathway for initialization and regridding.

kbarnhart:
@mandli - sorry for the delay. 

Yes, our `set_aux` skips filling/computing values if the value is not equal to `NEEDS_TO_BE_SET` (for example, [this line](https://code.usgs.gov/claw/dclaw/-/blob/add_rain/src/2d/dig/setaux.f90?ref_type=heads#L224) on the in development branch that needs this feature).

And I agree, some internal consistency is necessary. I understand the internal consistency with `set_aux`, so I've added a comment there. I have less familiarity with `icall`, so I am not entirely certain of the right comment to add with respect to `fillval`. 

If you have a suggestion, I can update the branch with the comment.

PR #733, opened 2026-08-26
Draft: Small stepgrid change

kbarnhart:
Another change based on D-Claw development: In D-Claw we have one variable in q that should not be set to zero when q(1)<drytol. 

This suggestion is intended to retain geoclaw functionality, but not require a D-Claw specific routine that differs only by this line. 

@rjleveque @mandli @dlgeorge


mandli:
This is probably true for any field that is momentum related I suppose.  Most of the other fields, unless they somehow have velocity issues, would not necessarily want to zero it out either.  In the multilayer code [we do the following](https://github.com/clawpack/geoclaw/blob/7ef7a8162c6466b58f416736ae663594dcb7bd54/src/2d/shallow/multilayer/stepgrid.f#L234):
```fortran
      forall(i=1:mitot, j=1:mjtot, k=1:num_layers,
     &       q(3*(k-1)+1,i,j) / rho(k) < dry_tolerance(k))
        q(3*(k-1)+1,i,j) = max(q(3*(k-1)+1,i,j), 0.d0)
        q(3*(k-1)+2,i,j) = 0.d0
        q(3*(k-1)+3,i,j) = 0.d0
      end forall
```
Given the momentum equations are interspersed with the other fields, this also required another `stepgrid.f`.

In any case, modulo the bouss code, I think being explicit about this is probably fine.  The alternative would be to call something that would do only this custom bit.  I suppose this might be a recurring task that may be easier to implement that way, but that may be too complex a solution for this.

kbarnhart:
@mandli - 

I think it be worth having `stepgrid` call a custom bit here would make sense. If Geoclaw, D-Claw, multi-layer all have custom `stepgrid` that differs only in these lines, reducing duplication seems worth it to me. 

e.g., for dclaw it would look like this:

```fortran

subroutine fixdry(q, nvar, mitot, mjtot)
      use geoclaw_module, only dry_tolerance
      use digclaw_module, only: i_h,i_hu,i_hv,i_hm,i_pb,i_hchi

      ! Input parameters
      integer, intent(in) :: nvar,mitot,mjtot

      ! Output
      double precision, intent(inout) :: q(nvar,mitot,mjtot)

      forall(i=1:my, j=1:mx, q(i_h,i,j) < dry_tolerance)
        q(i_h,i,j) = max(q(i_h,i,j),0.d0)
        q(i_hu,i,j) = 0.d0
        q(i_hv,i,j) = 0.d0
        q(i_hm,i,j) = 0.d0
        q(i_pb,i,j) = 0.d0
        q(i_hchi,i,j) = 0.d0
        ! do not reset i_bdif
      end forall
end subroutine fixdry
```

And for geoclaw it would look like:

```fortran
subroutine fixdry(q, nvar, mitot, mjtot)
      use geoclaw_module, only dry_tolerance

      ! Input parameters
      integer, intent(in) :: nvar,mitot,mjtot

      ! Output
      double precision, intent(inout) :: q(nvar,mitot,mjtot)

    forall(i=1:mitot, j=1:mjtot, q(1,i,j) < dry_tolerance)
        q(1,i,j) = max(q(1,i,j),0.d0)
        q(2:nvar,i,j) = 0.d0
      end forall

end subroutine fixdry
```

I can implement this in Geoclaw (new function, updates to the Makefile) if you think it would make sense. (suggestions for the subroutine name are welcome).

mandli:
@rjleveque and I chatted about this and were recalling that this type of operation is done a couple of places so having a `fixdry` subroutine that could be called from `stepgrid` and `b4step` would be fine.  My only thought is to change the signature so that we have:
```fortran
 ! =====================================================
 subroutine fixdry(meqn, mbc, mx, my, q, maux, aux)
 ! =====================================================
 !
 ! Reset cells whose depth is below dry_tolerance to a consistent dry state:
 ! zero negative depths to zero and zero the momenta.  Operates over the full
 ! patch, ghost cells included.
 !
 ! This routine is an override point.  Applications and GeoClaw variants may
 ! shadow it by placing their own fixdry.f90 in the application SOURCES list
 ! (see the GeoClaw docs on replacing library routines).  Two constraints on
 ! any replacement:
 !   * It must stay a plain external subroutine, not a module procedure
 !   * It must be thread safe.  Both call sites run inside an !$OMP PARALLEL DO
 !     over grids

     use geoclaw_module, only: dry_tolerance

     implicit none

     ! Arguments
     integer, intent(in) :: meqn, mbc, mx, my, maux
     real(kind=8), intent(inout) :: q(meqn, 1-mbc:mx+mbc, 1-mbc:my+mbc)
     real(kind=8), intent(in) :: aux(maux, 1-mbc:mx+mbc, 1-mbc:my+mbc)

     ! Locals
     integer :: i, j

     do j = 1-mbc, my+mbc
         do i = 1-mbc, mx+mbc
             if (q(1,i,j) < dry_tolerance) then
                 q(1,i,j) = max(q(1,i,j), 0.d0)
                 q(2,i,j) = 0.d0
                 q(3,i,j) = 0.d0
             end if
         end do
     end do

 end subroutine fixdry
```
The big difference is the addition of `aux` in case that is needed.

A couple of things to perhaps note:
- Since this is a call per-patch it should carry very little overhead (< 1%).  I think using `do` loops may be fine here and avoids a possible logical array creation (although I have never quite understood when this does and does not happen).
- There are other places where this happens but may not be so easy to plugin.  I had claude find some places where it may be worth a thought:
     1. src/2d/shallow/upbnd.f:120-122 — val(:,icrse,jcrse) = 0.d0 zeroes every component including h, contradicting the comment above it ("Reset small h to zeros") and skipping the max(h,0) every other site uses. For D-Claw this destroys i_bdif, the one component #733 exists to preserve. multilayer/upbnd.f:101-115 is the same code and additionally imports the scalar geoclaw_module::dry_tolerance and tests val(1,...) = ρ₁h₁ against it — off by ρ₁.
    2. src/2d/shallow/update.f90:168-170 — fine→coarse averaging writes only iadd(1..3). Components ≥4 on a coarse cell keep pre-update values forever (stale, not zeroed — worse to debug).
    3. src/2d/shallow/coarsen.f:109-111 — writes only valbgc(1..3); components ≥4 are fed uninitialized to the Richardson estimator via prepbigstep.f:84. Only reachable when flag_richardson = .true., which is False in every in-tree setrun.py.
    4. filpatch.f90:368 (do n = 2, nvar) and filval.f90:267 (do ivar = 2,nvar) — treat all of 2:nvar as h-scaled momentum: divide by h, velocity-bound against neighbours, remultiply by h_fine. Correct for hm/hchi, meaningless for a pressure like pb.
    5. prepregstep.f:24 and prepbigstep.f:88 — pass mitot/mjtot (resp. nx/ny against a mi2tot-sized array) as mx/my to b4step2, so it addresses out of bounds. fixdry inherits this verbatim; it is where a -fcheck=bounds build will trip. Unreachable today (flag_richardson = False everywhere in tree).
    6. amr2.f90 setprob ordering — setprob() runs before set_geo/set_storm/set_multilayer on fresh start (:521 vs :524/:531/:532) but after on restart (:486 vs :470/:471), so the existing procedure-pointer hooks cannot reliably be assigned by an app. A no-op setprob_final() stub in COMMON_SOURCES, called at the end of both branches, fixes it in ~8 lines.

kbarnhart:
@mandli @rjleveque - I've now updated this with Kyle's suggested function, put it into `stepgrid`, updated the Geoclaw `Makefile`, and tested it against both the Geoclaw and D-Claw bowl-slosh example.

I didn't think it made sense for me to replace it in other locations in Geoclaw, but if one of you wants to do that on this branch, you should have push access. 

Let me know if there are any other improvements to make or alternative approaches to implementation to discuss.

kbarnhart:
Working on some larger simulations and this (or something else I did) has seemed to make things slower. 

This is mostly to say that I think a timing test on full probably makes sense to do for this one. 

I can plan to do that (sometime this week at the latest, I hope) and report back.

kbarnhart:
OK, probably was something else. On a small (but not super small) problem, I get the following timing.

259.063 s CPU time on current master, 259.480 s with this change.

Without this change: 
```
============================== Timing Data ==============================

Integration Time (stepgrid + BC + overhead)
Level           Wall Time (seconds)    CPU Time (seconds)   Total Cell Updates
  1                    11.991                 22.318            0.463E+08
  2                    61.271                206.457            0.283E+09
total                  73.262                228.775            0.329E+09

All levels:
stepgrid               72.460                227.093    
BC/ghost cells          0.480                  1.349
Regridding             13.876                 20.281  
Output (valout)         1.337                  0.721  

Total time:            97.658                259.063  
Using  4 thread(s)

Note: The CPU times are summed over all threads.
      Total time includes more than the subroutines listed above
Note: timings are also recorded for each output step
      in the file timing.csv.
```


With this change

```
============================== Timing Data ==============================

Integration Time (stepgrid + BC + overhead)
Level           Wall Time (seconds)    CPU Time (seconds)   Total Cell Updates
  1                    12.009                 22.359            0.463E+08
  2                    61.310                206.685            0.283E+09
total                  73.319                229.044            0.329E+09

All levels:
stepgrid               72.512                227.349    
BC/ghost cells          0.484                  1.361
Regridding             13.901                 20.309  
Output (valout)         1.743                  0.725  

Total time:            98.357                259.480  
Using  4 thread(s)

Note: The CPU times are summed over all threads.
      Total time includes more than the subroutines listed above
Note: timings are also recorded for each output step
      in the file timing.csv.
```

mandli:
@kbarnhart I've pushed a commit finishing this off, sent as https://github.com/kbarnhart/geoclaw/pull/1 rather than straight to your branch so you get a look at it first — it makes a couple of design calls so take a look and let me know what  you think.  Summary below is specifics on what Claude found with me with regards to some of those design decisions and the performance question.

---

Summary here so the discussion stays on this thread (assisted by Claude).

**What it does.** Migrates the four remaining inline copies (`shallow/b4step2.f90`, `multilayer/stepgrid.f`, `multilayer/b4step2.f90`, `bouss/stepgrid.f`) and adds the variant routines `multilayer/fixdry.f90` and `bouss/fixdry.f90`, with source entries in `Makefile.multilayer` and `Makefile.bouss` — both are standalone lists, so each needed its own.

**The rule it encodes:** clamp the depth positive, zero the momenta, touch nothing else. Zeroing a component past the momenta destroys conservation of whatever it represents; if such a component needs handling at the dry state the right treatment is a positivity clamp like the depth's, not a zero. So the library routine stops at `q(2:3)` and leaves the rest to the application — which is what D-Claw needs for `i_bdif`. The reasoning is in the header so override authors get it along with the signature. Multilayer applies the same rule per layer. Bouss also zeros `q(4:5)`, but that is ψ, a momentum *rate* applied as `q(2:3) += dt*q(4:5)`, so it is momentum-like and covered by the same rule rather than an exception to it.

**On the timing question — it is faster, not slower.** Your instinct to check was right, for a reason neither of us guessed: gfortran lowers a mask-guarded `FORALL` into a `malloc` + `memset` + multiple passes + `free` of a `mitot*mjtot` mask, *per call*. The `do`+`if` form allocates nothing. 135 vs 55 instructions generated. Measured on the loop body alone, compiled exactly as production compiles it: 3.95x faster at `-O2`, 4.07x at `-O2 -fopenmp`, 5.07x at `-O3 -march=native`, 1.49x at `-O0` (the clawpack default, since `FFLAGS ?=` is empty). The call itself costs 0.0004 ns/cell against ~0.25 ns for the loop.

End-to-end on chile2010, alternating ABAB, n=15 per arm, `stepgrid` timer: 0.9885 at one thread, 95% CI [0.9835, 0.9914], p = 0.001 — a 1.15% speedup. At four threads it is 0.9878 but the CI is [0.957, 1.023], i.e. not resolvable. That last point is worth flagging for anyone repeating this: running master against *itself* at four threads gives a ratio of 1.007 with a CI of [0.940, 1.061], so this machine can't resolve anything below roughly 6% at that thread count. An effect this size only becomes visible in the one-thread arm, where the OpenMP scheduling noise is gone — which is why I ran both.

**One deliberate behaviour change.** `multilayer/b4step2.f90` zeroed the layer momenta but never clamped a negative layer depth, while `multilayer/stepgrid.f` did. Unifying gives `b4step2` the clamp. Instrumenting that call site on `bowl-radial`: 0 of 30,927,577 dry-cell visits would have clamped, so it closes a divergence without changing an in-tree result. Output is bit-identical on both multilayer examples.

**Verification.** chile2010 bit-identical to master across all 60 output files with `output_format = binary64`, so raw IEEE-754 rather than rounded ASCII — for `meqn = 3` this is pure code motion and the gate holds it to that. Same check on `plane_wave` and `bowl-radial`. `pytest` on the base and multilayer examples: 26 passed, 1 xfailed.

**Bouss is compile-verified only** — it needs PETSc, which I don't have working here. `bouss/fixdry.f90` and the edited `bouss/stepgrid.f` compile and the symbols wire up, but `examples/bouss/radial_flat` has not been run. @rjleveque, could you?

**A consequence worth recording.** With the fixup moved out, `multilayer/stepgrid.f` no longer differs from `shallow/stepgrid.f` in anything that matters. The entire remaining diff is a dead `use geoclaw_module` (dead in the shallow copy too now), `fgout_interp_needed` as an automatic array vs an allocatable, two `#ifdef WHERE_AM_I` debug blocks, and comment style. Pointing `Makefile.multilayer` at `$(GEOLIB)/stepgrid.f` and deleting the copy gives bit-identical output on both multilayer examples. Caveat: neither exercises `fgout`, which is exactly where the automatic-vs-allocatable difference lives, so that path is untested. I'll put that in a follow-up stacked on this rather than growing this PR further. `bouss/stepgrid.f` genuinely stays — it computes `eta` for the Boussinesq solve.

**Not included**, but found along the way and worth their own issues: `shallow/upbnd.f:119-122` zeros `val(:,...)` including the depth, contradicting its own comment and skipping the `max(h,0)` — for D-Claw that destroys `i_bdif`, the exact component this PR exists to protect. `multilayer/upbnd.f:8` tests `rho(1)*h_1` against the scalar `geoclaw_module::dry_tolerance`, off by `rho(1)`. `update.f90:168-170` and `coarsen.f:109-111` write only components 1..3, leaving 4+ stale. And `prepbigstep.f:88` passes 12 actual arguments to a 14-argument `b4step2`, while `prepregstep.f:24` passes `mitot,mjtot` as `mx,my` — both unreachable today since `flag_richardson` is False everywhere in tree, but `fixdry` inherits the bad bounds, so a `-fcheck=bounds` build will trip there.

Assisted-by: claude claude-opus-5[1m]

kbarnhart:
@mandli - I have merged your PR into my branch. 

Would you be willing to share your Claude prompts so I can learn from how you gave direction?

mandli:
@kbarnhart great!  Definitely willing to share.  Will follow up with you directly.

For others, the maintainers of Clawpack in general are working towards an AI policy, some of which may have been demonstrated over the past few months by me.  Stay tuned for when we release a draft and solicit general feedback!

PR #738, opened 2026-09-04
Small fgout tools fix

kbarnhart:
The following variables were never defined in the initial expansion of this code to support additional D-Claw variables. I'm now using fgout with additional variables so have discovered that these lines were missing. 

tag: @mandli @rjleveque @dlgeorge


rjleveque:
This looks fine, although I think `self._hm` is now defined twice.

kbarnhart:
good catch @rjleveque . I've now de-duplicated that.

PR #740, opened 2026-09-05
WIP - New example chile2010_netcdf not working

rjleveque:
I'm experimenting with this new `chile2010_netcdf` example, where I tried specifying

```
    topo_path = '/Users/rjl/git/AGTwork/topo/GEBCO_2026.nc'
    topo = topotools.Topography()
    topo.path = topo_path
    topo.topo_type = 4  # netCDF
    topo.crop_extent = [-190, -60, -60, 0]  # computational domain
    topo.buffer = 1  # buffer to make sure domain is covered
    topo.coarsen = 20 # coarsen from 30 arcsec to 10 arcmin
    topo_data.topofiles.append(topo)
```

to test using a crop extent that crosses the date line at longitude -180.

Before PR #739, the `topo.data` file only listed one topo grid, and would only get part of the `crop_extent`.

With this PR, `topo.data` lists the same `.nc` file twice, as desired, so two topo grids are created, as expected, plus a third one internally corresponding to the `dtopo` data file (`topo_for_dtopo`).

However, the dtopo is not longer applied, apparently -- no tsunami appears.   Cutting the domain and `crop_extent` back to starting at `x_lower = -120`, it works fine with the netCDF file as topo data, and the expected tsunami is generated.

Another issue: Even though I have `topo.buffer = 1` when GeoClaw runs it reports that the topo does not cover the full domain.  Again it works fine with `x_lower = -120`.  The problem I think is that there is a gap between the topo files at -180 that is hard to avoid with 2 files, since the GEBCO data starts at `x = -179.99791666666667` and ends at `x = 179.99791666666667`.  But I thought that buffering was perhaps supposed to deal with this case too?


mandli:
Just want to clarify the input data so I can replicate:
- topo is the newest version of the GEBCO full globe data set?
- dtopo is the same as the existing example?

rjleveque:
@mandl - correct!

PR #742, opened 2026-09-07
Fix `topo0save` bug with topo ordering interaction with dtopo

mandli:
In an effort to diagnose the problem in #740 another bug became apparent where `topo0save` indexing would not allow a `dtopo` file to reach every `topo` file it overlaps.  This appears in v5.14.0 from commit ebe27727b2a653f3d82cd1328fc4992dfdc0da05 that attempted to allow overriding the topo order.  Tests have been added to catch this.

Note that this does not yet solve the problem with #740.

### DIFF
--- a/src/2d/shallow/topo_module.f90
+++ b/src/2d/shallow/topo_module.f90
@@ -439,16 +439,27 @@ contains
                     call read_topo_file(mxtopo(i),mytopo(i),itopotype(i),topofname(i), &
                         xlowtopo(i),ylowtopo(i),topowork(i0topo(i):i0topo(i)+mtopo(i)-1), i)
                     ! set topo0save(i) = 1 if this topo file intersects any
-                    ! dtopo file.  This approach to setting topo0save is changed from 
+                    ! dtopo file.  This approach to setting topo0save is changed from
                     ! v5.4.1, where it only checked if some dtopo point lies within the
                     ! topo grid, which might not happen for small scale topo
-                    do j=mtopofiles - num_dtopo + 1, mtopofiles
-                        if ((xhitopo(i)<xlowtopo(j)) .or. &
-                            (xlowtopo(i)>xhitopo(j)) .or. &
-                            (yhitopo(i)<ylowtopo(j)) .or. &
-                            (ylowtopo(i)>yhitopo(j))) then
-                              topo0save(i) = 0
-                          else
+                    !
+                    ! The topo_for_dtopo entries occupy slots mtopofiles+1 ...
+                    ! mtopofiles+num_dtopo (filled above); mtopofiles itself is the
+                    ! real file count here, having been decremented after the
+                    ! allocates and not yet re-incremented.  Iterating
+                    ! "mtopofiles-num_dtopo+1, mtopofiles" instead compared the topo
+                    ! files against each other, which is only harmless when there is
+                    ! a single topo file (it then compares file 1 with itself).
+                    !
+                    ! Accumulate rather than assign: with num_dtopo > 1 an
+                    ! if/else that also cleared topo0save let each dtopo file
+                    ! overwrite the previous one's verdict, so only the last
+                    ! counted.  topo0save was zeroed above, so only ever set it.
+                    do j=mtopofiles + 1, mtopofiles + num_dtopo
+                        if (.not. ((xhitopo(i)<xlowtopo(j)) .or. &
+                                   (xlowtopo(i)>xhitopo(j)) .or. &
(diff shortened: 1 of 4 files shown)


mandli:

## The indexing error

`read_dtopo_settings` fills the `topo_for_dtopo` entries into slots
`mtopofiles+1 … mtopofiles+num_dtopo` (`topo_module.f90:407`), and reads them
back at `:480`. The `topo0save` loop at `:445` was instead written as

```fortran
do j = mtopofiles - num_dtopo + 1, mtopofiles
```

That range *was* correct when it was written (`be784494`, 2018): back then
`mtopofiles` had been incremented by `num_dtopo` and stayed that way, so the
last `num_dtopo` slots really were the dtopo entries. `ebe27727` split that
into an increment/decrement pair — `+num_dtopo` at `:196` for the allocates,
`-num_dtopo` back at `:250`, re-incremented only later at `:508`. Line 445
sits between the decrement and the re-increment, so `mtopofiles` there is the
**real file count** and the loop walks the last few *regular topo files*
instead of the dtopo entries.

The consequence is in `topo_update.f90:88`:

```fortran
if (topo0save(mt) <= 0) then
   topotime(mt) = t
   cycle          ! never updated -- this file's seafloor never moves
```

## Why it stayed hidden

With one topo file the range degenerates to `j = 1, 1`, so file 1 is compared
against **itself**. Self-overlap is always true, `topo0save(1) = 1`, and the
right answer comes out for the wrong reason. Every single-topo-file example —
including `chile2010` — is immune.

It needs two or more topo files to bite, and then only for files that do not
overlap whichever file the loop happens to land on.

## Second, latent bug in the same loop

The body was an `if/else` that set `topo0save(i) = 0` on no-overlap. Since the
loop runs over dtopo files, each iteration **overwrote** the previous one's
verdict, so with `num_dtopo > 1` only the last dtopo file counted. `topo0save`
is already zeroed at `:405`, so the fix only ever sets it — accumulating
across dtopo files rather than assigning.

## Test design

`tests/regression/topo0save/` runs the real `xgeoclaw` on two disjoint topo
files plus a dtopo, parameterized over which half holds the dtopo. Two
conditions are load-bearing and both are easy to get wrong — the first draft of
this test passed against the unfixed code:

1. **The topo files must be finer than the dtopo.** GeoClaw's internal
   `topo_for_dtopo` grid matches the dtopo, is unconditionally marked
   `topo0save = 1`, and is inserted into the priority order by cell area. If it
   is the finest thing over the footprint it supplies deformed seafloor
   regardless of what `topo0save` says about the real files, and the bug is
   invisible. Here topo is `0.02°` against a `0.1°` dtopo.
2. **The dtopo must be instantaneous** (`mt = 1`). A dtopo that ramps over an
   interval gets frozen at whatever fraction the first timestep reached — the
   earlier draft measured exactly 10% of the uplift — which couples the
   assertion to time-stepping for no benefit.

The two files are disjoint in longitude, mirroring the antimeridian wrap-split
that produced the multi-file `topo.data` in the first place, with a seam narrow
enough to stay inside the 1% domain-coverage tolerance (a wider gap is a hard
`stop` in `topo_module.f90`).

Parameterizing over both halves matters: only one half fails, so a
single-orientation test had even odds of missing this.

## Evidence

Mean bathymetry over the footprint interior, flat `-1000 m` seafloor, `+5 m`
uplift. Rebuild confirmed fresh in both directions (object mtime checked
against source mtime — `make new` does a full clean).

| | `west` | `east` |
|---|---|---|
| before fix | **-1000.0** (no deformation) | -995.0 ✓ |
| after fix | -995.0 ✓ | -995.0 ✓ |

`east` is the self-compared file, correct either way; `west` is the one the
loop marks as needing no update.

## Verification

- `tests/regression/topo0save/` — 4 passed with the fix; the two `west` cases
  fail without it.
- `examples/tsunami/chile2010/test_chile2010.py` — 6 passed. This is the
  single-topo-file path that the bug never affected, pinning that the fix
  changes nothing there.
- `pytest tests/ -m "not remote"` — 519 passed, 4 skipped, 1 xfailed.

## On #740

As the description says, this does not fix it. Worth recording *why*, so it is
not re-tried: the reported case coarsens GEBCO 20× to `0.167°`, while
`dtopo_usgs100227.tt3` is `0.101°`. The dtopo is finer, so `topo_for_dtopo`
outranks the topo entries over the source region and is deformed regardless of
`topo0save` — condition 1 above is not met there. The missing tsunami in #740
has a different cause and still needs reproduction against the original GEBCO
file.

PR #744, opened 2026-09-07
Fix bug in Isaac's `setplot.py` regarding loading storm data

mandli:
Fixes a bug that was introduced while testing additional met forcing things that assumed a storm format and availability.  The default case now works again and should now behave correctly while still providing example code to support various different use cases.

### DIFF
--- a/examples/storm-surge/isaac/setplot.py
+++ b/examples/storm-surge/isaac/setplot.py
@@ -44,7 +44,20 @@ def setplot(plotdata=None):
     friction_data.read(os.path.join(plotdata.outdir, 'friction.data'))
 
     # Load storm data
-    storm = stormtools.Storm(met_data.storm_file, file_format="data")
+    #
+    # The storm file's format follows the forcing family the run actually used,
+    # so read it off surge.data rather than hard-coding one: "data" is the
+    # gridded descriptor format, and feeding it a parametric track file makes
+    # GriddedMetForcing.read_data fail on a line it cannot parse.  storm_family
+    # is one of "parametric", "gridded" or "none" (see forcing_subtype_registry
+    # in geoclaw.data).
+    if met_data.storm_family == "gridded":
+        storm = stormtools.Storm(met_data.storm_file, file_format="data")
+    elif met_data.storm_family == "parametric":
+        storm = stormtools.Storm(met_data.storm_file, file_format="geoclaw")
+    else:
+        # Forcing is off: there is no storm file to read.
+        storm = None
     track = met_plot.track_data(os.path.join(plotdata.outdir, 'fort.track'))
 
     # Set afteraxes function


mandli:

## What broke

`setplot.py` hard-coded the storm file's format:

```python
storm = stormtools.Storm(met_data.storm_file, file_format="data")
```

`"data"` is the **gridded** descriptor format. But `setrun.py:431` uses
`storm_specification_type = 'holland80'` and writes the track with
`file_format='geoclaw'` (`setrun.py:442`), so the plotter handed a parametric
track file to `GriddedMetForcing.read_data`, which reads the first line as an
integer and gets an empty string:

```
File ".../geoclaw/met/gridded.py", line 154, in read_data
    self.file_format = int(data_file.readline().partition("#")[0].rstrip())
ValueError: invalid literal for int() with base 10: ''
```

The hard-coded value only ever matched the commented-out
`storm_specification_type = 'data'` branch, so the example's own default could
not plot. isaac was the only setplot in the tree passing `file_format="data"`.

## The change

Select the format from the family the run actually recorded.
`SurgeData.read` populates `storm_family` from `surge.data`
(`geoclaw/data.py:1054`), and it is one of `"parametric"`, `"gridded"` or
`"none"` per `forcing_subtype_registry` (`geoclaw/data.py:856`):

| `storm_family` | `file_format` |
|---|---|
| `parametric` (holland80, willoughby, CLE, …) | `geoclaw` |
| `gridded` (OWI / NetCDF) | `data` |
| `none` | no file is read; `storm = None` |

Reading it off `surge.data` rather than off `setrun.py` means the plotter
follows whatever was actually run, including a run whose setrun has since been
edited.

The `none` case is newly handled. Previously, forcing turned off meant there
was no storm file and the unconditional load raised.

## Why the storm object is kept

`storm` is currently passed to `storm_gauge_afteraxes(cd, storm)` and not read
there — that function plots from `met_plot.gauge_wind(cd)` and
`met_plot.gauge_pressure(cd)`, i.e. from the gauge solution rather than the
track. Dropping the load would therefore also have fixed the crash and been a
smaller diff.

It is deliberately retained: a loaded `Storm` exposes track position and
intensity, which is genuinely useful to have in hand when extending the gauge
plots, and this example is where people look for how to get one. Keeping it
also means the failure mode is fixed rather than side-stepped — it now works
for both forcing families instead of one.

## Verification

Ran the example's own `setrun` for both configurations (its ATCF source
`bal092012.dat` is local, so no download is involved) and exercised the exact
call that was failing:

| `storm_specification_type` | `storm_family` | derived format | this PR | old hard-coded `"data"` |
|---|---|---|---|---|
| `holland80` (the default) | `parametric` | `geoclaw` | OK | `ValueError: invalid literal for int() with base 10: ''` |
| `data` | `gridded` | `data` | OK | OK |

The right-hand column reproduces the reported traceback exactly, and shows the
old code worked only for the non-default configuration.

Not run: `test_isaac.py`, whose cases are marked `remote` and download
topography, and which exercise the simulation rather than `setplot`. Checked by
inspection instead: `setplot_kml.py` loads no `Storm`, and the remaining
`file_format="data"` occurrences (the `setrun.py` write path and
`test_isaac.py`) are all inside gridded-specific branches and are correct.

PR #727, opened 2026-08-13
Combine `buffer` and anti-meridional wrapping

mandli:
Previously using a `buffer` for a topography file was limited to not use the `TopoInspector` functionality, notably the anti-meridional wrapping (domain cross data-line).  This modifies the code so that Python understands how to do this with a buffer.

Assisted-by: claude opus-4-8

PR #748, opened 2026-09-07
Add topo/dtopo input roadmap to dev/design

mandli:
Design document for tracking and future plans for the heavy redo of the input file formats, notably `topo`, `dtopo` and `met` input.

### DIFF
--- a/dev/design/README.md
+++ b/dev/design/README.md
@@ -44,3 +44,7 @@ lost. Several of these docs *were* lost that way (they lived in a gitignored
   verification plan (**\*\*partial reconstruction\*\***; original lost, no full mirror).
 - `friction-file-plan.md` — file-based friction field input implementation plan
   (relocated from `.plans/`).
+- `topo\_input\_roadmap.md` — living roadmap for the topo/dtopo input rework:
+  merge order for the linear PR stack, superseded branches and what replaced
+  them, and the deferred items (antimeridian seam gap, dtopo Fortran parity,
+  1D preprocessing scope, build-flag hygiene).
--- a/dev/design/topo_input_roadmap.md
+++ b/dev/design/topo_input_roadmap.md
@@ -4,11 +4,14 @@
 rework, what has landed, what was superseded and why, and what is deliberately
 deferred.*
 
-**Provenance.** Written 2026-09-07 from the branch and PR state at that date.
-Every claim below was checked against the repositories rather than from notes;
-where a claim is *not* verified it says so. Companion to
-`met_forcing_roadmap.md`, which covers the meteorological side of the same
-input-handling effort.
+**Provenance.** Written 2026-09-07 from the branch and PR state at that date,
+and amended later the same day after test-merging the open PRs in order: §1
+(the stack forks at #741; #745 rebased onto it), §2.3 (layout guards, and the
+`make .output` trap), §4 (#745's verification re-anchored to its post-rebase
+commits) and §5 (new — why the sequence broke). Every claim below was checked
+against the repositories rather than from notes; where a claim is *not* verified
+it says so. Companion to `met_forcing_roadmap.md`, which covers the
+meteorological side of the same input-handling effort.
 
 ---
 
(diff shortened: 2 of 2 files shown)


mandli:

## Why this is tracked rather than in `.plans/`

`dev/design/README.md` makes the case already: a source of truth in a gitignored
path cannot be reviewed, cannot be updated in the same PR as the code, cannot be
`git blame`d, and can be lost — several met-forcing docs *were* lost that way.
This roadmap is the topo/dtopo equivalent, and until now there was no topo
design doc at all; only the met ones existed.

The Contents entry in `dev/design/README.md` is updated in the same commit,
which that file's own same-PR rule requires.

## What it records

**§1 — merge order.** The topo stack is strictly linear (#739 → #741 →
`topo-input-parity`), verified with `git merge-base --is-ancestor` rather than
assumed, plus the five independent PRs and the geoclaw-before-clawutil ordering
for #745 / clawutil#208.

**§1 — supersession, with the evidence inline.** #727 and `topo-input-unify` are
being closed and deleted, so the reasoning has to survive them:

- #727's mechanism was *reimplemented*, not dropped — a four-row table maps each
  piece (`_expand_wrapped_topo` → `_resolve_topo_records`; Python-side buffer
  baking → Fortran `nbuf4 = topo_buffer(topo_idx)`) and each of its three tests
  to a named, stronger equivalent in #739.
- `topo-input-unify`'s six phase commits are listed against their 1:1 rebased
  equivalents on parity (`19573e4e→994e6efb`, and so on).

This is what makes the close/delete auditable after the fact. It also records
that `c893f238` *fixed* a bug in the same area: leaving `nbuf4 = 0` on the
`nc_has_crop` branch silently dropped `topo_buffer` for every file
`topo_entries()` writes. `master` still has the single assignment; #739 has both.

**§2 — five deferred items**, each with what blocks it rather than a bare TODO:

| | |
|---|---|
| ±180 seam gap | design decision; three specific questions written out |
| dtopo Fortran crop/wrap parity | mandli/geoclaw#15 |
| 1D preprocessing scope | opened by #745; needs a decision, not a default |
| build-flag hygiene | `radial_flat` hardcode; the `ALL_LFLAGS` question |
| `-DNETCDF` / `make new` | documented, with both rejected fixes recorded |

The seam-gap entry is the one worth reading: it explains why `buffer` *cannot*
close the hole (after #739 the buffer is applied, but clamped to the file's own
extent, and neither entry has data past its edge), so the next person does not
re-try tuning it.

**§3 — the #740 correction on record.** The `topo0save` bug in #742 was
initially believed to explain the missing tsunami. It does not: #740 coarsens
GEBCO 20× to `0.167°` while `dtopo_usgs100227.tt3` is `0.101°` (read from the
file header), so the finer `topo_for_dtopo` grid outranks the topo entries and
is deformed regardless. Recorded so the theory is not re-tried.

**§4 — how each claim was checked**, including what was *not*: that parity
passes CI on a machine other than the author's, and the `remote` tests.

## The CI correction in the second commit

The first version of §2.3 said CI selects only `python` and `slow` markers.
That is wrong — `testing.yml` also has a `regression-tests` job, and
`slow-tests.yml` has matrix `regression`/`python` jobs. Found while checking
whether parity's `tests/regression/topo_crop/` suite would actually run in CI.

The conclusion was unaffected and still holds — `git grep -l 1d_classic --
tests/ .github/` is empty, so **no 1D example is built or run anywhere in CI** —
but the supporting detail would have misled anyone auditing coverage, and it
mattered in the other direction too: `regression` is a first-class CI selector,
which is what #742's `topo0save` test and parity's `topo_crop` suite rely on.

The paragraph now carries a table of all five jobs with their marker strings
quoted verbatim from the workflow files, each checked against
`grep -hoE '\-m "[^"]+"' .github/workflows/*.yml`.

## Verification

Documentation only, so the check is that the references are real. Each was
confirmed against the repositories at 2026-09-07:

- `a3ee4b27` (#726 merge), the two `merge-base` ancestry relations, parity's
  12 commits.
- `nbuf4 = topo_buffer` occurrence counts: `master` 1, #739 2.
- `Makefile.common:66` / `:84`, `examples/bouss/radial_flat/Makefile:76`.
- `dtopo_usgs100227.tt3` `dx = 1.01010101010104e-01` from its header.
- mandli/geoclaw#15 open, with the title quoted.
- All five CI marker strings verbatim; `1d_classic` absent from `tests/` and
  `.github/`.

PR #749, opened 2026-09-07
Unify topo and dtopo input handling: shared crop core, coordinate tools, antimeridian gate

mandli:
This is the last piece of the topo/dtopo input work, and the point of it is to have **one** implementation of "read a window out of a gridded file" rather than the two that had drifted apart between topo and dtopo. This will hopefully make it easier to intuit what capabilities exist and how they work as well as help with maintainability.

Two new modules carry the shared machinery:
- `coordinate_tools.py` — the crop/coarsen/buffer/align index math (`crop_indices`) plus the longitude-axis classification used to decide whether a file may wrap across the anti-meridian. This is the routine #726 introduced for topo and is now combined.
- `gridded_input.py` — a small format-reader for the different types of files (`ASCIIReader`, `NetCDFReader`, `GeoTIFFReader`) replacing the `topo_type` if/elif ladder in `topotools.py`. These are thin wrappers over the existing read functions so no new I/O, but cleans up seams in the implementation and allows easy extensions to new formats such as zarr.  Note that this does not actually implement GeoTIFF beyond leveraging existing capabilities in the shared path.

`dtopo` then gets the same treatment as `topo`: `crop_extent`, `coarsen`, `buffer` and `align` on read, applied across every time slice of `dZ`, and `DTopoInspector.dtopo_entries()` gains the anti-meridian wrap-split and descriptor `crop_bounds`. That is what "parity" in the branch name refers to. The parity is not finished as it still needs the contents of mandli/geoclaw#15 for full parity, but is waiting on merging the rest of this first.

## Caution - Coordinate System Assumptions

Anti-meridian wrapping is now gated on `geodata.coordinate_system` instead of being guessed per file, and a mismatch is a hard error in **both** directions: a geographic file under a Cartesian run (`= 1`) raises, and a projected file under a lon-lat run (`= 2`) raises. Fortran enforces the same thing when reading the descriptor, for topo and dtopo alike.  This was deliberate as mixing coordinate systems could error subtly.  If `coordinate_system` is not available via the calling sequence the per-file guess is still in place so old examples should still work.

## Merge order

This branch contains #739 and #741, so those merge first; until they do the PR shows all 12 commits rather than just the 7 that are new here.

It also conflicts with #745 in `TopographyData.write()` — #745 added the 1D branch to the old write loop, and this branch replaced that loop with `_resolve_topo_records`. I'll rebase once #745 lands. Every other open branch (#742, #743, #744, #748) merges clean.

## Still open

- Fortran dtopo crop/wrap parity — mandli/geoclaw#15
- The ±180 seam gap for cell-centered global DEMs — see `dev/design/topo_input_roadmap.md` §2.1; this branch does not close it, and the roadmap explains why `buffer` structurally cannot.


mandli:

Mechanics behind the description above: what each commit does, the API surface,
and how it was verified.

## The seven commits

The first five of the twelve are #739 and #741. These are the new ones:

| Commit | | Net |
|---|---|---|
| `994e6efb` | Extract `coordinate_tools` (`is_geographic_lon`, lon-entry geometry) out of `netcdf_utils` | +214 / −90 |
| `16538959` | Phase 1 — shared `crop_indices` + the `gridded_input` reader registry for topo | +663 / −310 |
| `8a631fa1` | Phase 2 — gate antimeridian wrapping on `coordinate_system` (Python + Fortran) | +291 / −18 |
| `de4cdb8c` | Phase 3a — dtopo crop/coarsen/buffer/align on read via the shared `crop_indices` | +183 / −20 |
| `6f9969bf` | Phase 3b — `DTopoInspector.dtopo_entries` wrap-split + gate; dtopo descriptor `crop_bounds` | +154 |
| `64988a92` | Phase 3c — gate dtopo wrapping on `coordinate_system` (Fortran read) | +17 |
| `286ad84d` | Reconcile the shared crop core with #739's two-axis `crop_extent` guards | +189 / −14 |

`16538959` is the one to read first: it is where `topotools.py` loses ~310 lines
of format dispatch to `gridded_input`, so its diff looks large but is mostly
relocation.

## API surface

`coordinate_tools.py`:

| | |
|---|---|
| `crop_indices(coords, lo, hi, delta, coarsen, buffer, align)` | the single index-window computation |
| `axis_file_slice(...)` | file-order slice, including descending-latitude storage |
| `classify_lon_axis` / `is_geographic_lon` | is this x-axis a longitude? |
| `resolve_wrap(coordinate_system, nature)` | may it wrap? |
| `_compute_lon_entries` | the wrap-split geometry |
| `EmptyCropWindow(ValueError)` | crop selects nothing |

`gridded_input.py` is a registry: `register_reader()` maps each reader's
`topo_types` into `_REGISTRY`, `get_reader(topo_type)` dispatches on
`abs(topo_type)` (the sign only flips the elevation convention, which stays with
the caller), and readers implement one method, `read_window()`, returning a
`GridData`. Unit/fill/preprocessing handling deliberately stays in the product
rather than moving into the adapters.

## The gate, precisely

`resolve_wrap` has three outcomes, and the third is what keeps existing setups
working:

| `coordinate_system` | geographic file | projected file |
|---|---|---|
| `1` (Cartesian) | `ValueError` | no wrap |
| `2` (lon-lat) | wrap | `ValueError` |
| `None` (not wired through) | wrap | no wrap — legacy per-file heuristic, no cross-check |

Fortran mirrors this at descriptor-read time for both topo and dtopo: a non-zero
`lon_wrap_offset` under `coordinate_system == 1` prints the offending file and
`stop`s, rather than silently applying a meaningless offset.

## dtopo

`DTopography._apply_crop()` reuses `coordinate_tools.crop_indices` for both axes
and applies the window across every time slice — `dZ[:, jlo:jhi:coarsen,
ilo:ihi:coarsen]` — so the deformation stays consistent in time. It short-circuits
when nothing is requested, and raises naming the file when `crop_extent` does not
overlap it. `read_header()` is also added, matching the topo lazy-load pattern.

## Verification

Run as CI runs it — from the repo root with `-o addopts=""`, so the
example-based tests are collected too, not just `tests/`:

| Selection | Result |
|---|---|
| `python and not remote` | 449 passed, 4 skipped, 3 xfailed |
| `regression and not slow and not adjoint and not remote` | 38 passed, 2 xfailed (110 s) |

The regression run builds `xgeoclaw` and includes this branch's own
`tests/regression/topo_crop/` suite, which drives the real binary across the
seam — that is the coverage most likely to catch a mistake here, so it was the
one to insist on.

Parity adds ~2400 lines of test, including `tests/test_coordinate_tools.py`
(new), `tests/test_gridded_input.py` (new), `tests/test_topo_data_golden.py`
(new, with five golden `topo.data` files), and a 611-line extension to
`tests/test_topotools_preprocessing.py`.

**Not verified here:** anything marked `remote` (skipped throughout), and CI on
any platform other than macOS/arm64 — the ubuntu `python` jobs on this PR have
since passed; the `regression` jobs were still running when this was written.

## Conflict check

Every open branch was checked with `git merge-tree` against the merge base
rather than assumed:

| vs | conflicts |
|---|---|
| #742 `fix-topo0save-dtopo-index` | 0 — hunks at 442-452 against 294/1504/1777 in `topo_module.f90`, no overlap |
| #743 `fix-multilayer-openmp-link` | 0 |
| #744 `fix-isaac-setplot-storm-format` | 0 |
| #748 `topo-input-roadmap` | 0 |
| #745 `fix-1d-topo-data-format` | **3**, all in `TopographyData.write()` |

The #745 conflicts are genuine rather than incidental: this branch replaced the
write loop with `_resolve_topo_records`, #745 added the 1D branch inside the old
loop, and both introduce a module-level helper at the same point in the file
(`_reject_remote_path` vs `_warn_unsupported_1d_preprocessing` — that third one
is trivial to resolve).

PR #752, opened 2026-09-12
Fix off-by-one in force_dry mask indexing

### LINKED ISSUE #751: force_dry mask is indexed off by one: qinit.f90 shifts it in x, filpatch/filval shift it in y
## Summary

The `force_dry` lookup appears in five places in the 2d shallow/bouss code, and they do
not agree with each other. `qinit.f90` is off by one column in x; `filpatch.f90` and
`filval.f90` (both the `shallow` and the `bouss` copies) are off by one row in y. Since
all five compute the same thing from the same arrays, at most one convention can be right,
and in fact each site is correct on one axis and wrong on the other.

Affected: v5.14.0, and still present on master (`11479f67`, 2026-01-24).

## The five call sites

`src/2d/shallow/qinit.f90` (lines 44-51):

```fortran
do i=1,mx
    x = xlower + (i-0.5d0)*dx
    ii = int((x - xlow_fdry + 1d-7) / dx_fdry)
    do j=1,my
        y = ylower + (j-0.5d0)*dy
        jj = int((y - ylow_fdry + 1d-7) / dy_fdry)
        jj = my_fdry - jj  ! since index 1 corresponds to north edge
```

`src/2d/shallow/filpatch.f90` (338-340), `src/2d/shallow/filval.f90` (235-237), and the
identical `src/2d/bouss/filpatch.f90` and `src/2d/bouss/filval.f90`:

```fortran
ii = int((xcent_fine - xlow_fdry + 1d-7) / dx_fdry) + 1
jj = int((ycent_fine - ylow_fdry + 1d-7) / dy_fdry) + 1
jj = my_fdry - jj  ! since index 1 corresponds to north edge
```

Note that `qinit.f90` has no `+ 1` on either line, while the other four have `+ 1` on both.

## Why both are wrong

`force_dry` is allocated `(mx_fdry, my_fdry)`, i.e. 1-based, in
`src/2d/shallow/qinit_module.f90:292`, and row index 1 is the north row. These lookups only
run when the patch resolution matches the mask resolution (the `ddxy` test in `qinit.f90`,
`use_force_dry_this_level` elsewhere), so cell centers and mask cells are aligned.

For a cell center in mask column `k`, counted 1-based from the west:

    int((x - xlow_fdry + 1d-7) / dx_fdry) == k - 1

so the correct column index is `int(...) + 1`. `qinit.f90` omits it and reads column `k-1`,
the western neighbour.

For a cell center in mask row `m`, counted 1-based from the south, `int(...) == m - 1`. The
array is stored north row first, so the correct row index is

    my_fdry - m + 1 == my_fdry - int(...)

which is what `qinit.f90` computes. The other four compute `my_fdry - (int(...) + 1)`,
i.e. one row too far.

So the correct pair, at all five sites, is:

```fortran
ii = int((x - xlow_fdry + 1d-7) / dx_fdry) + 1
jj = my_fdry - int((y - ylow_fdry + 1d-7) / dy_fdry)
```

This does not depend on whether `xlow_fdry`/`ylow_fdry` are read as the corner of the mask
raster or as the center of its first cell: on an aligned grid the two conventions change the
quotient by exactly 0.5, and `int()` gives the same result for both.

## Effect

The mask is applied one cell away from where it was computed, so cells that should start dry
start wet and vice versa. It is easy to miss because the shift is small and silent: the
out-of-range index at the edge (`ii = 0` in `qinit.f90`, `jj = 0` on the north row in
`filpatch`/`filval`) is filtered by the surrounding `(ii>=1) .and. ...` test, so there is no
error message and no out-of-bounds access, and the whole westernmost column (resp. the
northernmost row) is simply never forced dry.

## Suggested fix

Apply the two lines above at all five sites.

alorentiar:
The force_dry lookups were using two different conventions. qinit.f90
was reading one column too far west (no +1 on the column index), while
filpatch/filval in both the shallow and bouss copies were reading one row
too far north (an extra +1 before the north-edge flip). So the mask lands
one cell away from where it was computed, and the wrong cells start dry
or wet.

Made all five sites use the same convention:

    ii = int((x - xlow_fdry + 1d-7) / dx_fdry) + 1
    jj = my_fdry - int((y - ylow_fdry + 1d-7) / dy_fdry)

Fixes #751.


mandli:
Thanks for this, and welcome to the community!  This is definitely something I think that needs to be fixed and I confirmed the convention independently and wrote up the details on #751.

## Some Comments

A few things before it goes in:

**1. `int()` should be `floor()`.** Your change fixes the mask interior. It leaves a second off-by-one at the west and south edges, because Fortran `int()` truncates toward zero rather than flooring, so a cell just outside the mask produces a quotient in `(-1, 0)` and lands on column 1 or row `my_fdry` instead of failing the bounds test. Numbers are in my comment on #751. In `qinit.f90` the x case is newly exposed by adding the `+1`, so it is worth folding into this PR:

```fortran
ii = floor((x - xlow_fdry + 1d-7) / dx_fdry) + 1
jj = my_fdry - floor((y - ylow_fdry + 1d-7) / dy_fdry)
```

**2. Verification.** This is a results-affecting change to a feature with no test coverage at all, so the green CI here carries no information: `grep -rn force_dry tests/` is empty and the one example is never built by CI. I wrote a regression test for it (flat bathymetry, an asymmetric mask, an exact dry/wet map compared against an expectation computed in Python from the same mask, no stored golden). Four Fortran cases cover `qinit.f90` at `t0`, `qinit` on a refined patch, and `filval` on a patch created by a later regrid, plus two pure-Python guards on the file contract and on parity across the five sites. Results:

| build | outcome |
|---|---|
| master | all 6 fail |
| with this PR | the full-domain-mask case passes; the three cases involving a mask smaller than the domain still fail |
| with this PR plus `floor()` | all 6 pass |

So your change fixes the shift completely. What it leaves is the halo in (1), and it shows up in three of the four Fortran cases, `filval` included. Concretely, on a 4x4 all-ones mask on a 10x8 domain, this branch dries 9 cells outside the mask rectangle: the whole column immediately west of it and the whole row immediately south. I will link the branch here; you are welcome to pull it into this PR so the fix and its test land together.

**3. The `src/2d/bouss/` copies.** Those cannot be exercised by CI, since the bouss build needs PETSc and MPI and we do not install either in the workflows. Did you check them against the shallow versions, or apply the same edit on the assumption that they match? They do match, byte for byte in that block, so the change is right either way, but I would rather the PR said which it was.

**4. Optional, and not a condition of merging.** Five hand-maintained copies of the same index expression drifting apart is how this bug happened. If you wanted to go further, hoisting the map into a single helper in `qinit_module` (something like `force_dry_index(x, y, ii, jj)`) would collapse all five sites to one call. I am happy to take the narrow fix as it stands if you would rather not widen the scope.

Assisted-by: claude claude-opus-5

---

**One other thing.** Clawpack does not have an AI contribution policy yet. We are working on one, and I am not going to apply a rule that does not exist. But I would like to ask directly: were this PR, or the issue it fixes, drafted with the help of an AI tool? If so, we would like it on the record. The low-friction form is a commit trailer:

```
git commit -s --trailer "Assisted-by: Claude Code (<model>)"
```

To be explicit about why, because I know this question can land as hostile and it is not meant that way: disclosure tells a reviewer where to look, it is never held against a contributor, and the answer does not change whether this gets merged. The `-s` sign-off matters on its own terms, since only a person can certify a contribution and that is the part a tool cannot do for you.

I hold my own AI-assisted PRs in this repo to the same thing.

What I would most like to see is that whoever is submitting this can stand behind it. That is really what the verification in (2) is for, and why I would rather you wrote or adapted the test than have me merge it around you. Replies here in your own words please, rather than routed back through a model.

Thanks again for finding this. It is a genuine bug that has been in here since ~2020.

PR #753, opened 2026-09-14
Units policy: ASCII coverage and one override mechanism

mandli:
This closes the two holes left (purposefully) in the `UNITS_POLICY` that is include din #741 for topo and dtopo ASCII files.

Stacked on #749 (which carries #739 and #741).


mandli:
Stacked on #749 (which carries #739 and #741). One commit of its own: `8155d73b`.

> **Base branch note.** `units-ascii-coverage` is currently the only branch of this stack pushed to `clawpack/geoclaw`; the others live on the fork. If `topo-input-parity` is also pushed here, this PR's base can be set to it and the diff collapses to the single commit above. Otherwise the GitHub diff shows all 13 commits of the stack and only the last one is new.

## Summary

#741 wrote GeoClaw's units policy down and made it enforceable, then deliberately left two rows open — marked `gap=` in `UNITS_POLICY` and `xfail(strict=True)` in the suite:

- `topo_ascii` — ASCII carries no units and no override; elevation in cm or feet read as meters, with no message and no sanity check.
- `dtopo_ascii` — the same, for ASCII dtopo.

Both are now closed. **Every row of the generated table reads "yes" under Conforms**, and the conformance suite goes from 15 passed / 2 xfailed to 29 passed / 0 xfailed.

## The hoist is what made ASCII cheap

`gridded_input`'s own module docstring says "all shared post-processing (unit / fill / preprocessing handling) stays in the product", and `Topography.read` comments its reader call site the same way — but unit resolution, conversion and the magnitude check all still lived *inside* `NetCDFReader.read_window`. So this finishes what #749 started rather than adding a second copy to `ASCIIReader`:

- `GridData` gains `source_units` and `var_name`. Readers now **report** what the file declared; `None` means the format cannot declare anything at all.
- `Topography._resolve_units` does the rest once, for every format: resolve (declared → `assume_units` → contract), convert, magnitude-check.

ASCII support then falls out. The first run after the hoist flipped `topo_ascii` to `XPASS(strict)` with no ASCII-specific code written at all.

### Two ordering bugs the hoist surfaced

I first placed the shared step before missing-data normalization. That was wrong twice over, and the second way is the one worth knowing about:

1. The magnitude check saw the raw `no_data_value` sentinel and rejected a perfectly good file (`[-99999, 5] m`). Visible, easy to catch.
2. Unit conversion would have scaled the sentinel too — a km file's `-99999` becoming `-99999000` — so it would no longer match `self.no_data_value`, would never be normalized to NaN, and every missing cell would silently become a −1e8 m seafloor. Invisible, and exactly the class of bug this PR series exists to remove.

The step now runs after normalization, with the reasoning recorded in a comment so it does not get "tidied" back later.

## `assume_units` is now a real argument

It previously lived in `nc_params`, a NetCDF-specific channel whose name is meaningless for a `.tt3`:

- `Topography.read(..., assume_units=None, skip_sanity_check=None)`. The `nc_params` spellings still work; supplying both raises rather than silently picking one.
- `DTopography.read(..., assume_units=None)`, forwarded to `DTopoInspector` for `dtopo_type=4` and applied directly for the ASCII types.
- `MetInspector.assume_units` accepts a `{role: unit}` mapping. `True` still works behind a `DeprecationWarning`.

On that last point: the `bool` was never a mistake. Met covers several variables with *different* contract units, so a single string would be ambiguous — which is why `format_units` already existed in mapping form. But the bool can only assert "already in contract units", so it cannot convert, and it never said which variable it was meant to cover. The mapping does both, and `units_policy.md` records why the type differs from the other inspectors rather than leaving it to look like an inconsistency.

`Topography.__init__` forwards `**kwargs` to `read()`, so `Topography(path, skip_sanity_check=True)` works for free — a consequence of promoting the argument instead of burying it in a dict.

## The two ASCII rows differ from each other, and the table shows it

- **topo ASCII → `assume+check`.** Meters is assumed, and rule 5 (the magnitude check) catches a gross error such as a file in centimeters.
- **dtopo ASCII → `warn+assume`.** `_check_magnitude` defines **no bound for deformation**, so there is no rule-5 backstop here. With nothing to catch a mistake, the assumption is announced instead. Passing `assume_units='m'` states it deliberately and silences the warning.

I did not invent a deformation bound to make the two match. Setting one is a judgement about coseismic displacement physics rather than about code, so it is recorded as open in `units_policy.md`. `'assume+check'` is a new vocabulary value precisely so this distinction is visible, instead of both rows claiming a protection only one of them has.

## Wider recognition, without weakening rule 4

Unit matching was exact-string and case-sensitive, so `Meters` and `KM` raised, and `ft`/`feet` had no entry at all — meaning GeoClaw demanded a units declaration and then refused a correct one. Now:

- Lookup is case-folded. The tables stay explicit; only the lookup relaxes.
- Feet added to `units.py` and `_CF_TO_UNITS_PY`. Checked against a known value: 29032 ft → 8849 m.
- `furlongs`, `m^2`, `degrees` and `''` still raise, pinned by `test_case_folding_did_not_weaken_rule_4`.
- **The CF alias tables' British spellings are untouched.** `'metre'`, `'metres'`, `'metres per second'` and `'kilometre'` are *data* — real NetCDF files declare them, and "correcting" them to US spelling would silently break reading those files. Pinned by `test_cf_alias_data_survived_case_folding`.

## Fallout worth knowing about

Turning the magnitude check on for ASCII trips **four existing tests**, all reading `topo_bowl` / `topo_bowl_hill`: analytic parabolas peaking at 17 km and 9.25 km. They now pass `skip_sanity_check=True`, with a comment explaining why.

I checked whether this reaches the examples before accepting it, and it does not. The real `bowl-slosh` example uses `h0 = 0.1` over `[-2, 2]`, giving a range of ±0.7 m — nowhere near the bound. The 17 km peak is an artifact of the *test's* deliberately wide domain, not of idealized topography in general. So the exposure is four test fixtures, not the example suite.

This is still a behavior change and wants a release note: an ASCII topo file whose elevations fall outside roughly [−11000, 9000] m now raises instead of being read silently. `skip_sanity_check=True` is the documented escape hatch for synthetic-but-valid data.

## Verification

- **The conformance suite is the verification**, and both rows flipped via `XPASS(strict)`, which forced their registry updates. That loop caught me twice during this work — it is doing its job.
- **Round-trip rather than a plausible run:** a `.tt3` written in cm and read with `assume_units='cm'` equals the same data written in meters, to 1e-9.
- **11 of the new tests fail when the behavior files are reverted.** The three that pass either way are the guards — rule-4 preservation, CF-alias data integrity, and `nc_params` back-compat — which is correct for guards, since they pin things that must not change.
- **The hoist is behavior-preserving for NetCDF:** the `met_forcing` and `topo_crop` regression goldens, the `topo.data` goldens, and the `*_test_data.tt3` dtopo baselines are all byte-identical.
- **607 passed, 4 skipped, 1 xfailed**; 12 regression tests.

## Documentation

- `units_policy.md`: the ASCII section's "it is only *assumed*" caveat is now wrong and is replaced with what actually happens, including why the two ASCII rows differ. The generated table is regenerated from the registry.
- `dev/design/README.md` gains the Contents entry for `units_policy.md` that **#741 should have added** — that directory's own same-PR rule requires it, and it was missed.
- The roadmap's §2.6 entry covering this work is on #748 separately, because `topo_input_roadmap.md` does not exist on this branch and so cannot be edited from here.

## Still open

- **A magnitude bound for deformation**, which would let dtopo ASCII reach `assume+check` alongside topo ASCII.
- **`FrictionData.write`** writes `"'%s' %s\n " % fname` — two format specifiers, one argument — so it raises `TypeError` whenever `friction_files` is non-empty. It has zero test coverage, and the feature is a known-dormant stub (`friction_module.f90` prints "File based friction specification unimplemented") with its own tracked design doc. Recorded, deliberately unfixed here.

Assisted-by: claude claude-opus-5[1m]

